PR #43, opened 2024-11-10
Better example for generating poems.

madiator:
We show four examples of how to generate poems.
1. Generate a single poem in a single call.
2. Generate many poems in a single call.
3. Generate n poems by making n calls.
4. Lastly, to show how we can improve diversity, we generate topics and then generate poem one by one for each topic.


RyanMarten (approved):
LGTM!

Issue #46, opened 2024-11-10
[UI] Resolve dataset viewer component streaming logic after PR #28


CharlieJCJ:

vutrung96:
Fixed with #79

Issue #45, opened 2024-11-10
[UI] Resolve dataset viewer component read file location after PR #28


CharlieJCJ:

vutrung96:
Fixed with #79

PR #28, opened 2024-11-08
Add OpenAIBatch backend and refactor RequestProcessor to be compatible

RyanMarten:
Now you can use batch via 
```
curator.Prompter(
    prompt_func=prompt_func, 
    parse_func=parse_func, 
    model_name="gpt-4o-mini", 
    batch=True
)
```

New example in `examples/distill.py` which shows this for a large reannotation job. 

Also includes small change to the interface:
prompt_func returns a dictionary
```{ "user_prompt": content, "system_prompt": content}```
or
```{ "user_prompt": content: "system_prompt": content}```
to
prompt_func returns a list (of standard form messages or string 
`content`
which gets converted to
```
[
 {"role": "user", "content": content}
]
```
or the messages
```
[
 {"role": "system", "content": content}
 {"role": "user", "content": content},
]
```

PR #49, opened 2024-11-10
[Hot fix] Fixed file path for batch=False prompter requests

CharlieJCJ:
#28 introduces a new cache storing logic, causing frontend cache loading behaviors to change. 

In this PR, addresses for #45 

For `batch=False` requests, there's a rename to the `responses.jsonl` -> `responses_0.jsonl` with the same streaming logic (check for lastline). 

**The data viewer is expected to behave normally after this PR for all `batch=False` requests, including `examples/poem.py` and `examples/camel.py`.** 

Here is a quick fix for #45, there will be a separate PR that deals with batch=True requests, in which a separate streaming logic will be implemented that address #46

### DIFF
--- a/bespoke-dataset-viewer/app/api/responses/[runHash]/route.ts
+++ b/bespoke-dataset-viewer/app/api/responses/[runHash]/route.ts
@@ -12,7 +12,7 @@ export async function GET(
 ): Promise<Response> {  // This is the key change
   try {
     const { runHash } = await params
-    const responsesPath = join(homedir(), '.cache', 'curator', runHash, 'responses.jsonl')
+    const responsesPath = join(homedir(), '.cache', 'curator', runHash, 'responses_0.jsonl')
     
     const searchParams = request.nextUrl.searchParams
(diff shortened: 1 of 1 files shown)


madiator:
Why not `responses_nobatch.jsonl`?

vutrung96:
Technically online requests can also have batches: https://github.com/bespokelabsai/curator/blob/a475722f0f724f651d65d3ffc31a037518d199cd/src/bespokelabs/curator/request_processor/openai_online_request_processor.py#L31.

@CharlieJCJ I think we should just make a general fix for multiple batches here. IIUC once we have that it should work for both online and batch.

CharlieJCJ:
yeah, this PR was meant for fixing the current `batch=False` cases since batch parameter in online isn't been used https://github.com/bespokelabsai/curator/blob/main/src/bespokelabs/curator/prompter/prompter.py#L74. So just as a temporary fix such that ppl can view the UI on non-batch scenario. My next PR is built on this. 

Two solutions: 
- merge this, and I quickly follow another PR on this
- close this, I only have one PR

PR #53, opened 2024-11-11
Alow specifiying working directory in case users want to use a different working directory for their cache

vutrung96:
DCFT has a concrete need where separate outputs from different operators need to be stored in different cache directories.

I also feel like this would be a useful feature for users generally (huggingface datasets has a similar way of specifying file names in https://huggingface.co/docs/datasets/v3.1.0/en/package_reference/main_classes#datasets.Dataset.map.cache_file_name)

### DIFF
--- a/src/bespokelabs/curator/prompter/prompter.py
+++ b/src/bespokelabs/curator/prompter/prompter.py
@@ -22,6 +22,7 @@ from bespokelabs.curator.request_processor.openai_online_request_processor impor
     OpenAIOnlineRequestProcessor,
 )
 
+_CURATOR_DEFAULT_CACHE_DIR = "~/.cache/curator"
 T = TypeVar("T")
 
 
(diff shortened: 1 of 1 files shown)


src/bespokelabs/curator/prompter/prompter.py
```diff
@@ -104,9 +111,12 @@ def _completions(
         if self is None:
             raise ValueError("Prompter must be provided")
 
-        curator_cache_dir = os.environ.get(
-            "CURATOR_CACHE_DIR", os.path.expanduser("~/.cache/curator")
-        )
+        if working_dir is None:
+            curator_cache_dir = os.environ.get(
+                "CURATOR_CACHE_DIR", os.path.expanduser("~/.cache/curator")
```
madiator, line 116: ~/.cache/curator should be a global variable at the top of the file.

_CURATOR_DEFAULT_CACHE_DIR = "~/.cache/curator"

RyanMarten (approved):
LGTM!

Issue #54, opened 2024-11-11
Catch TypeError when using TikToken to count tokens


RyanMarten:
Error trace: 
```
  File "/Users/ryan/dcft_private/.venv/lib/python3.12/site-packages/bespokelabs/curator/prompter/prompter.py", line 78, in __call__
    return self._completions(self._request_processor, dataset)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ryan/dcft_private/.venv/lib/python3.12/site-packages/bespokelabs/curator/prompter/prompter.py", line 159, in _completions
    dataset = request_processor.run(
              ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ryan/dcft_private/.venv/lib/python3.12/site-packages/bespokelabs/curator/request_processor/openai_online_request_processor.py", line 183, in run
    asyncio.run(
  File "/opt/homebrew/Cellar/python@3.12/3.12.6/Frameworks/Python.framework/Versions/3.12/lib/python3.12/asyncio/runners.py", line 194, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.12/3.12.6/Frameworks/Python.framework/Versions/3.12/lib/python3.12/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.12/3.12.6/Frameworks/Python.framework/Versions/3.12/lib/python3.12/asyncio/base_events.py", line 687, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/Users/ryan/dcft_private/.venv/lib/python3.12/site-packages/bespokelabs/curator/request_processor/openai_online_request_processor.py", line 351, in process_api_requests_from_file
    token_consumption=num_tokens_consumed_from_request(
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ryan/dcft_private/.venv/lib/python3.12/site-packages/bespokelabs/curator/request_processor/openai_online_request_processor.py", line 677, in num_tokens_consumed_from_request
    num_tokens += len(encoding.encode(value))
                      ^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #59, opened 2024-11-12
Catch TypeError when using tiktoken and fall back to hueristic token counting

RyanMarten:
Addresses #54

### DIFF
--- a/src/bespokelabs/curator/request_processor/openai_online_request_processor.py
+++ b/src/bespokelabs/curator/request_processor/openai_online_request_processor.py
@@ -674,7 +674,15 @@ def num_tokens_consumed_from_request(
             for message in request_json["messages"]:
                 num_tokens += 4  # every message follows <im_start>{role/name}\n{content}<im_end>\n
                 for key, value in message.items():
-                    num_tokens += len(encoding.encode(value))
+                    try:
+                        num_tokens += len(encoding.encode(str(value)))
+                    except TypeError:
+                        logger.warning(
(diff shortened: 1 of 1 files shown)


src/bespokelabs/curator/request_processor/openai_online_request_processor.py
```diff
@@ -674,7 +674,15 @@ def num_tokens_consumed_from_request(
             for message in request_json["messages"]:
                 num_tokens += 4  # every message follows <im_start>{role/name}\n{content}<im_end>\n
                 for key, value in message.items():
-                    num_tokens += len(encoding.encode(value))
+                    try:
+                        num_tokens += len(encoding.encode(str(value)))
+                    except TypeError:
+                        logger.warning(
+                            f"Failed to encode value {value} with tiktoken to count tokens, instead assuming a token for every 4 characters."
```
madiator, line 681: count tokens, instead -> count tokens. Instead

src/bespokelabs/curator/request_processor/openai_online_request_processor.py
```diff
@@ -674,7 +674,15 @@ def num_tokens_consumed_from_request(
...
+                        )
+                        num_tokens += (
+                            len(str(value)) // 4
+                        )  # assume a token for every 4 characters
```
madiator, line 685: num_tokens += len(str(value)) // 4
(one line and no comment since you just explained it above).

madiator (approved):

PR #61, opened 2024-11-12
Prevent .arrow file getting in an invalid state and generate different .arrow based on parse_func

### LINKED ISSUE #40: Changing parse func should not invalidate the cache
A critical use of the cache is when parsing errors out, you still want the saved responses and not re-run generation.

### LINKED ISSUE #51: Prevent cache from getting invalid state when ArrowWriter fails
Catch exceptions when using Arrow writer and delete the file before raising the exception to the user, so the cache doesn't get into an invalid state (.arrow file is present but not readable).

This results in the following error: 

```
Traceback (most recent call last):
  File "/home/mahesh/code/curator/examples/poem.py", line 84, in <module>
    generate_diverse_poems(how_many=10)
  File "/home/mahesh/code/curator/examples/poem.py", line 76, in generate_diverse_poems
    poems = poet(topics_dataset)
            ^^^^^^^^^^^^^^^^^^^^
  File "/home/mahesh/code/curator/src/bespokelabs/curator/prompter/prompter.py", line 78, in __call__
    return self._completions(self._request_processor, dataset)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mahesh/code/curator/src/bespokelabs/curator/prompter/prompter.py", line 159, in _completions
    dataset = request_processor.run(
              ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mahesh/code/curator/src/bespokelabs/curator/request_processor/openai_online_request_processor.py", line 197, in run
    dataset = self.create_dataset_files(dataset, working_dir, prompt_formatter)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mahesh/code/curator/src/bespokelabs/curator/request_processor/base_request_processor.py", line 243, in create_dataset_files
    return Dataset.from_file(dataset_file)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mahesh/.cache/pypoetry/virtualenvs/bespokelabs-curator-tZ21X9LR-py3.11/lib/python3.11/site-packages/datasets/arrow_dataset.py", line 740, in from_file
    table = ArrowReader.read_table(filename, in_memory=in_memory)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mahesh/.cache/pypoetry/virtualenvs/bespokelabs-curator-tZ21X9LR-py3.11/lib/python3.11/site-packages/datasets/arrow_reader.py", line 329, in read_table
    return table_cls.from_file(filename)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mahesh/.cache/pypoetry/virtualenvs/bespokelabs-curator-tZ21X9LR-py3.11/lib/python3.11/site-packages/datasets/table.py", line 1017, in from_file
    table = _memory_mapped_arrow_table_from_file(filename)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mahesh/.cache/pypoetry/virtualenvs/bespokelabs-curator-tZ21X9LR-py3.11/lib/python3.11/site-packages/datasets/table.py", line 63, in _memory_mapped_arrow_table_from_file
    opened_stream = _memory_mapped_record_batch_reader_from_file(filename)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mahesh/.cache/pypoetry/virtualenvs/bespokelabs-curator-tZ21X9LR-py3.11/lib/python3.11/site-packages/datasets/table.py", line 49, in _memory_mapped_record_batch_reader_from_file
    return pa.ipc.open_stream(memory_mapped_stream)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mahesh/.cache/pypoetry/virtualenvs/bespokelabs-curator-tZ21X9LR-py3.11/lib/python3.11/site-packages/pyarrow/ipc.py", line 190, in open_stream
    return RecordBatchStreamReader(source, options=options,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mahesh/.cache/pypoetry/virtualenvs/bespokelabs-curator-tZ21X9LR-py3.11/lib/python3.11/site-packages/pyarrow/ipc.py", line 52, in __init__
    self._open(source, options=options, memory_pool=memory_pool)
  File "pyarrow/ipc.pxi", line 1006, in pyarrow.lib._RecordBatchStreamReader._open
  File "pyarrow/error.pxi", line 155, in pyarrow.lib.pyarrow_internal_check_status
  File "pyarrow/error.pxi", line 92, in pyarrow.lib.check_status
pyarrow.lib.ArrowInvalid: Tried reading schema message, was null or length 0
```

This happened since parse_func errored out and  didn't write properly to the dataset file in a previous run and this is trying to read the cached dataset.arrow. 

The solution is then to fix the parse_func bug, but then you get an arrow trying to read the invalid file from the previous run. 

Rather than requiring the user to manually manage the cache (deleting the .arrow file), we should prevent it from getting in this state.

RyanMarten:
# 🔧 Improve caching behavior for parse functions

## 🎯 Changes

* 🗃️ Moved `parse_func` hash out of cache folder naming
* 📝 Use `parse_func` hash as the `.arrow` dataset filename instead
* 🔄 This allows regenerating datasets from cached responses when only the parse function changes
* ⚠️ Added more helpful error messages for invalid parse functions
* 🧹 Added automatic cleanup of corrupted `.arrow` files to prevent cache getting stuck

## ✨ Benefits

Now you can:
- Iterate on your parse function without re-running expensive LLM calls
- Get clear guidance when your parse function returns invalid data
- Avoid getting stuck with corrupted cache files

Closes #51
Closes #40


src/bespokelabs/curator/prompter/prompter.py
```diff
@@ -113,13 +113,14 @@ def _completions(
         )
 
         prompt_func_hash = _get_function_hash(self.prompt_formatter.prompt_func)
+
+        # Used as the name of the dataset .arrow file so changing this will keep the same cache for the LLM responses, but will create a new dataset file
```
madiator, line 117: Line too long. Please split.
(Google convention is 80 characters per line for Python).

src/bespokelabs/curator/prompter/prompter.py
```diff
@@ -157,7 +158,10 @@ def _completions(
 
         # TODO(Ryan): do the response processing, while context of original dataset is available and need random access via row_idx)
         dataset = request_processor.run(
-            dataset, f"{curator_cache_dir}/{fingerprint}", self.prompt_formatter
+            dataset,
+            f"{curator_cache_dir}/{fingerprint}",
```
madiator, line 162: os.path.join(curator_cache_dir, fingerprint) is safer.

src/bespokelabs/curator/request_processor/base_request_processor.py
```diff
@@ -223,10 +229,17 @@ def create_dataset_files(
...
             raise ValueError(f"No responses files found in {working_dir}")
-        dataset_file = f"{working_dir}/dataset.arrow"
+        dataset_file = f"{working_dir}/{parse_func_hash}.arrow"
         if os.path.exists(dataset_file):
             logger.info(f"Using existing dataset file {dataset_file}")
-            return Dataset.from_file(dataset_file)
+            try:
+                output_dataset = Dataset.from_file(dataset_file)
+                return output_dataset
+            except Exception as e:
```
madiator, line 238: We shouldn't catch broad exceptions. It can mask many issues.
For example if there was a type (return otput_dataset), that will throw an exception which will be caught here.

src/bespokelabs/curator/request_processor/base_request_processor.py
```diff
@@ -275,8 +288,19 @@ def create_dataset_files(
             try:
                 writer.finalize()
             except SchemaInferenceError as e:
+                os.remove(dataset_file)
+                raise ValueError(
+                    "Arrow writer is complaining about the schema: likely your `parse_func` returned only None objects. Please fix your `parse_func` to return valid rows and re-run. The dataset will be regenerated from the cached LLM responses."
+                ) from e
+            except TypeError as e:
+                os.remove(dataset_file)
                 raise ValueError(
```
madiator, line 297: Let's still do TypeError?

src/bespokelabs/curator/request_processor/base_request_processor.py
```diff
@@ -275,8 +288,19 @@ def create_dataset_files(
...
-                    "Arrow writer is complaining about the schema: likely all of your parsed rows were None and writer.write only wrote None objects."
+                    "Arrow writer is complaining about the types: likely your `parse_func` is not returning a valid list of rows (list of dictionaries). Please fix your `parse_func` to return valid rows and re-run. The dataset will be regenerated from the cached LLM responses."
```
madiator, line 298: Split into multiple lines.

src/bespokelabs/curator/request_processor/base_request_processor.py
```diff
@@ -275,8 +288,19 @@ def create_dataset_files(
...
                 ) from e
 
-        return Dataset.from_file(dataset_file)
+        try:
+            output_dataset = Dataset.from_file(dataset_file)
+        except Exception as e:
```
madiator, line 303: another broad exception catching here here..

vutrung96: i think this is ok since we are not doing anything here and just forwarding the error in a more understandable manner?

we can say something like, f"Failed to load dataset from {dataset_file}. This was likely corrupted by a failed previous run. You can delete this file and try again to see if the error is transient."

src/bespokelabs/curator/request_processor/base_request_processor.py
vutrung96: Thinking about this more, we probably shouldn't remove the file here, but raise a ValueError similar to what I suggested in the comment below

 f"Failed to load dataset from {dataset_file}. This was likely corrupted by a failed previous run. You can delete this file and try again to see if the error is transient."

vutrung96 (changes requested):
Can we merge into dev instead of main since dev is our "staging" branch now?

src/bespokelabs/curator/request_processor/base_request_processor.py
RyanMarten: @madiator The user shouldn't have to know about the dataset file since it is in the cache. If you are using Prompter and it starts complaining about a dataset file, you would probably be confused and wouldn't know what to do right? At least adding an additional comment about the relation to parse_func helps put people on the track of fixing it.

src/bespokelabs/curator/request_processor/base_request_processor.py
RyanMarten: Changed it to catch a specific error here `pyarrow.lib.ArrowInvalid` instead. 
Removing the file is better IMO since this happens at the top of the logic - the user might have fixed their `parse_func` from last time, but not deleted the file  (we shouldn't expect or encourage users to touch the cache).

src/bespokelabs/curator/request_processor/base_request_processor.py
RyanMarten: I moved the return out of the try catch, but left the naked exception handling. 
I think since it is at the top of the logic, just deleting and retrying makes sense to me. If it fails again, the next `Dataset.from_file()` call only catches `pyarrow.lib.ArrowInvalid` so the  Exception will be surfaced again if there still is an issue. 

Also OK to change this naked exception to `pyarrow.lib.ArrowInvalid`, but just trying to help the user not touch the cache.

RyanMarten:
> Can we merge into dev instead of main since dev is our "staging" branch now?

Merging into `dev` now.

RyanMarten:
Actually is this all better handled upstream? (we can add this in addition)
We can have some asserts on the output of `parse_func` which check if valid rows are being returned and raise exceptions there.

RyanMarten:
Handled row validation upstream to give specific feedback to the user

vutrung96:
LGTM with handling validation upstream

vutrung96 (approved):

PR #63, opened 2024-11-12
Add logo

madiator:
Will update the starter example subsequently.


README.md
```diff
@@ -1,6 +1,27 @@
-# Bespoke Curator
+<p align="center">
+  <a href="https://bespokelabs.ai/" target="_blank">
+    <img src="./docs/Bespoke-Labs-Logo-on-Mint.png" width="50%" alt="Bespoke Labs Logo" />
+  </a>
+</p>
+
+<h1 align="center">Curator</h1>
```
CharlieJCJ, line 7: Thinking out loud here for alternative layout of logo and title. 

Ideally the text font can be consistent for bespoke labs (in logo) and curator IMO, 
right now it looks like this in my preview
image

something from my gut telling me having consistent font can be more appealing and just cleaner in general. 
image
Something like this is cleaner to me.

Alternative, we create a logo just for bespoke labs curator, including the text (example vllm)

Or alternatively, we don't have logo, and just put bespoke labs curator as title (example llama).

madiator: I didn't realize the effect on dark. Changed it now so that it works for both the themes.
Light theme:
image
Dark theme:
image

Also, I want to keep the Logo + Bespoke Labs at the top. 
And given we are small, we need one consistent logo, so new logo for Curator may not be a good idea.
We can definitely explore having Logo + Bespoke Labs Curator later.

madiator: Ok added just the logo. We can make Logo + Bespoke Labs Curator in photoshop later (Alex will try to do that).

madiator: done

CharlieJCJ: image
perfect

CharlieJCJ (approved):
LGTM!

PR #65, opened 2024-11-12
GenericRequest and GenericResponse refactor

### LINKED ISSUE #58: batch mode `requests_*.jsonl` files doesn't follow GenericRequest data schema
```
class GenericRequest(BaseModel):
    model: str
    messages: List[Dict[str, Any]]
    row: Dict[str, Any] | str
    row_idx: int
    response_format: Optional[Type[BaseModel]] = None
```

while in `batch=True` requests_0.jsonl (which includes in responses_0.jsonl) is 
```
(base) ➜  a2a86c77ae470e3f head requests_0.jsonl -n 1 | jq
{
  "custom_id": "0",
  "method": "POST",
  "url": "/v1/chat/completions",
  "body": {
    "model": "gpt-4o-mini",
    "messages": [
      {
        "role": "user",
        "content": "Write a very long, elaborate, descriptive and detailed shooting script, including a background and dialogues, for a John Waters-inspired film scene that includes a character preparing omelettes. You are free to choose the setting, scenario (it should make sense) and characters (give them names, and describe their appearance and clothing in detail)."
      }
    ]
  }
}
```

causing issues when access messages fields when doing `batch=True` frontend

RyanMarten:
Now:
- requests_N.jsonl contains a GenericRequest per line
- responses_N.jsonl contains a GenericResponse per line 
- RequestProcessor reads from the file converts from GenericRequest to api specific request on the fly
- When responses are received, the api specific response is converted to GenericResponse on the fly
- GenericResponse now contains GenericRequest so monitoring can get the original request information in a standard model 

Closes #58


tests/test_cache.py
vutrung96, line 1: why are adding an empty file?

diff-file.patch
vutrung96, line 1: remove this file too?

vutrung96 (approved):
LGTM

PR #67, opened 2024-11-12
Package versioning downgrade for colab

### LINKED ISSUE #30: Pypi package version is too new
image

CharlieJCJ:
Closes #30


CharlieJCJ:
> ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
gcsfs 2024.10.0 requires fsspec==2024.10.0, but you have fsspec 2024.9.0 which is incompatible.

this is a colab specific warning, I don't think we need to add additional package to revert this colab complaint. the other dependencies are installed correctly

vutrung96 (approved):

PR #69, opened 2024-11-12
Remove the use of asyncio.run to make asyncio work in colab

### LINKED ISSUE #68: Colab asyncio error after pip install
image

vutrung96:
We're using asyncio.run in our processor, but this will fail in colab where there is a loop already running since asyncio.run creates a new asyncio loop.

What we need to do is first checking if there are any running loops and use that loop instead. If a loop doesn't already exist, we create it.

Closes #68


RyanMarten (approved):
LGTM!

PR #70, opened 2024-11-12
Add Prompter arg for batch size

### LINKED ISSUE #42: Optional batch_size input variable for Prompter
So you can set this for the backend processor
Defaults to 1k
@CharlieJCJ and @vutrung96 agree on this

Don't allow batch_size for OnlineProcessor, only when batch=True

RyanMarten:
Closes #42 

batch_size is only used when batch=True


CharlieJCJ (approved):
LGTM, thanks!

PR #71, opened 2024-11-12
Cleanup build script and .gitignore for build artifacts

CharlieJCJ:
Changes:
- Added `/dist` to gitignore
- Have an exclude list for frontend folders that needs to be excluded from the `/static` folder


vutrung96 (approved):
did you test if the package works if we exclude .next?

CharlieJCJ:
Yes, I tested, there are post-build artifacts in .next that are useful, but not cache

PR #72, opened 2024-11-12
Fix asyncio with nest_asyncio

vutrung96:
nest_asyncio allows running nested loops by patching asyncio.


CharlieJCJ (approved):
LGTM, thanks for the fix @vutrung96 !

PR #73, opened 2024-11-12
Set RLIMIT_NOFILE to avoid "too many files open" errors

RyanMarten:
Sending thousands of requests in parallel from a single machine sometimes hits the soft resource limit for number open files.
This change increases this limit enough to accommodate the Prompter job.

### DIFF
--- a/src/bespokelabs/curator/request_processor/openai_online_request_processor.py
+++ b/src/bespokelabs/curator/request_processor/openai_online_request_processor.py
@@ -6,6 +6,7 @@ import re
 import time
 from dataclasses import dataclass, field
 from typing import Any, Callable, Dict, Optional, Set, Tuple, TypeVar
+import resource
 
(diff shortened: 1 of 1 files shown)


src/bespokelabs/curator/request_processor/openai_online_request_processor.py
```diff
@@ -211,6 +212,13 @@ async def process_api_requests_from_file(
         resume_no_retry: bool = False,
     ) -> None:
         """Processes API requests in parallel, throttling to stay under rate limits."""
+
+        # Increase the number of open file descriptors to avoid "Too many open files" errors
+        soft, hard = resource.getrlimit(resource.RLIMIT_NOFILE)
```
vutrung96, line 217: will this permanently change the limit for users?

vutrung96 (approved):

PR #75, opened 2024-11-12
Fix types in jsonl files

RyanMarten:
Was missing some json.loads statements so dicts were actually strings in GenericResponse


CharlieJCJ (approved):
LGTM, thanks for the fix @RyanMarten !

PR #84, opened 2024-11-13
Pre-emptively remove invalid dataset file when prompt_func detected as invalid

RyanMarten:
This was the warning is avoided unless something unexpected happens to corrupt the file otherwise.

### DIFF
--- a/src/bespokelabs/curator/request_processor/base_request_processor.py
+++ b/src/bespokelabs/curator/request_processor/base_request_processor.py
@@ -267,10 +267,17 @@ class BaseRequestProcessor(ABC):
 
                         # parse_func can return a single row or a list of rows
                         if prompt_formatter.parse_func:
-                            dataset_rows = prompt_formatter.parse_func(
-                                response.generic_request.original_row,
-                                response.response_message,
-                            )
+                            try:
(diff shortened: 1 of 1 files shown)


src/bespokelabs/curator/request_processor/base_request_processor.py
```diff
@@ -267,10 +267,17 @@ def create_dataset_files(
...
+                                dataset_rows = prompt_formatter.parse_func(
+                                    response.generic_request.original_row,
+                                    response.response_message,
+                                )
+                            except Exception as e:
+                                logger.error(
+                                    f"Exception raised in your `parse_func`. {error_help}"
+                                )
+                                os.remove(dataset_file)
+                                raise e
```
madiator, line 280: This is ok since you are reraising the exception.
Relevant snippet from Google's styleguide: https://google.github.io/styleguide/pyguide.html

Never use catch-all except: statements, or catch Exception or StandardError, unless you are

re-raising the exception, or
creating an isolation point in the program where exceptions are not propagated but are recorded and suppressed instead, such as protecting a thread from crashing by guarding its outermost block.

madiator (approved):

PR #85, opened 2024-11-13
Fix JSON parsing from model

vutrung96:
Since we are not using strict mode, it's possible for OpenAI to return invalid JSON (either not a JSON or valid JSON but not conform to the JSON schema). In this case we catch the invalid JSON and skip the response.

### DIFF
--- a/src/bespokelabs/curator/request_processor/base_request_processor.py
+++ b/src/bespokelabs/curator/request_processor/base_request_processor.py
@@ -307,7 +307,7 @@ class BaseRequestProcessor(ABC):
 
 def parse_response_message(
     response_message: str, response_format: Optional[BaseModel]
-) -> tuple[any, Optional[list[str]]]:
+) -> tuple[Optional[dict | str], Optional[list[str]]]:
     response_errors = None
     if response_format:
         try:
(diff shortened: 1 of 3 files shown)


src/bespokelabs/curator/request_processor/base_request_processor.py
```diff
@@ -259,11 +259,19 @@ def create_dataset_files(
...
+                                    )
                                 )
-                            )
+                            except ValidationError as e:
+                                logger.warning(
+                                    f"Pydantic failed to parse response message {response.response_message} with `response_format` {prompt_formatter.response_format}."
+                                    f"The model likely returned a JSON that does not match the schema of the `response_format`. Will skip this response."
+                                )
+                                response.response_message = None
+                                response.response_errors = [str(e)]
```
RyanMarten, line 274: Small nit: can we add our warning message here too so we can help the user understand the (confusing) pydantic error message

RyanMarten (changes requested):
Small change to the error message. 

Just say the model successfully responded with a string that is JSON but doesn't match the schema

RyanMarten (approved):
LGTM!

PR #97, opened 2024-11-14
Response format for batch bug fix

RyanMarten:
We are now reading the response format from Generic request as a dict, not a pydantic class. It is no longer necessary to get the json schema since it is already presented as such


CharlieJCJ (approved):
LGTM!

PR #106, opened 2024-11-14
Add an example for summarizing text messages between two people.

madiator:
- Generates diverse personas.
- Creates random pairings between personas with varying conversation parameters:
  - Conversation length (very short to very long)
  - Persona relatedness (very related to very unrelated)
- Generates synthetic text message conversations between paired personas
- Produces one-sentence summaries of the conversations.

### DIFF
--- a/examples/camel.py
+++ b/examples/camel.py
@@ -1,8 +1,5 @@
-import asyncio
 from typing import List
-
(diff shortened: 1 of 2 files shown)


RyanMarten:
@madiator if we want to keep this example, it needs to be updated to `curator.LLM`

PR #114, opened 2024-11-15
Terminal interactive usage of bespokelabs curator package

CharlieJCJ:
`inspect.getsource` wasn't working when used in interactive python. 

Here is a patch such that does another try except for OSError, to return empty as result of `inspect.getsource`, bypassing the terminal error. 

**We strongly encourage users to use bespoke curator in python scripts, for better cache tracking of `prompt_func` and `parse_func` and visualization.**

### DIFF
--- a/src/bespokelabs/curator/prompter/prompter.py
+++ b/src/bespokelabs/curator/prompter/prompter.py
@@ -180,11 +180,11 @@ class Prompter:
         metadata_db = MetadataDB(metadata_db_path)
 
         # Get the source code of the prompt function
-        prompt_func_source = inspect.getsource(
+        prompt_func_source = _get_function_source(
             self.prompt_formatter.prompt_func
         )
         if self.prompt_formatter.parse_func is not None:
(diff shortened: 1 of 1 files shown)


src/bespokelabs/curator/prompter/prompter.py
```diff
@@ -221,4 +221,12 @@ def _get_function_hash(func) -> str:
...
 
-    return xxh64(inspect.getsource(func)).hexdigest()
+    return xxh64(_get_function_source(func)).hexdigest()
+
+
+def _get_function_source(func) -> str:
+    """Get the source code of a function."""
+    try:
+        return inspect.getsource(func)
+    except OSError:
```
vutrung96, line 231: can you document why we're doing this in a comment?

vutrung96 (approved):

src/bespokelabs/curator/prompter/prompter.py
CharlieJCJ: yes!

PR #117, opened 2024-11-15
Update README and the poem example.

madiator:
* Add an improved example in the README that helps understand curator better.
* Remove lots of documentation that is in poem.py, which can be moved to our docs.


README.md
```diff
@@ -37,22 +37,41 @@ pip install bespokelabs-curator
...
+    prompt_func=lambda row: f"Write two poems about {row['topic']}.",
     model_name="gpt-4o-mini",
 )
 
-poem = poet()
-print(poem["response"][0])
+poem = poet(topics)
+print(poem.to_pandas())
 ```
+Note that `topics` can be created with `curator.Prompter` as well,
```
CharlieJCJ, line 65: I think we can also include an example print output from this example in README if ppl just want to see what happens. 

A GIF works well for a demo of curator-viewer in README. Can add later

README.md
```diff
@@ -37,22 +37,41 @@ pip install bespokelabs-curator
...
-import os
+from datasets import Dataset
+from pydantic import BaseModel, Field
+from typing import List
 
-os.environ['OPENAI_API_KEY'] = 'sk-...' # Set your OpenAI API key here
+# Create a dataset object for the topics you want to create the poems.
+topics = Dataset.from_list([{"topic": "Dreams vs. reality"},
+                            {"topic": "Urban loneliness in a bustling city"},
+                            {"topic": "Beauty of Bespoke Labs's Curator library"}])
```
CharlieJCJ, line 47: Change to
```python
topics = Dataset.from_dict({"topic": [
    "Dreams vs. reality",
    "Urban loneliness in a bustling city",
    "Beauty of Bespoke Labs's Curator library"
]})
```

Just a formatting nit, perhaps people like `from_dict` better? To illustrate it's a one column with three rows. Repeating "topic" key doesn't look good code, may give ppl bad impression.

README.md
madiator: Done

README.md
madiator: Done. I felt the from_list method is a bit clearer to show the rows created.
But this is fine.

CharlieJCJ:
Actually reviewing now 😄

CharlieJCJ:
image

Something fishy in UI display, I'll fix it

CharlieJCJ:
Ahh, need to escape this `\\n` instead of `\n`

CharlieJCJ:
image
image

Look like this now

PR #130, opened 2024-11-15
Add a teacher.py example that slightly cleans up Ryan's original code

vutrung96:


examples/teacher.py
```diff
@@ -0,0 +1,186 @@
...
+
+while len(needs_improvement_dataset) > 0 and iterations < max_iterations:
+    iterations += 1
+
+    # Step 1: Student generates responses for remaining instructions
+    response_dataset = student(needs_improvement_dataset)
+
+    # Step 2: Evaluator scores the responses
+    evaluation_dataset = evaluator(response_dataset)
+    # Step 3: Separate successful responses (score >= 9) from those needing improvement
```
CharlieJCJ, line 156: Can be a bit clearer in the comment, (all 4 criteria scores >= 9)

vutrung96: Good catch! fix.

PR #133, opened 2024-11-17
adding an env example file

lavishsaluja:


CharlieJCJ:
TODO here
- [ ] update __main__.py for `curator-viewer` to take environment variables if these exist (but I think perhaps the CLI defaults here are sufficient as well?)
- [ ] update README for `cp .env.example .env` operation and how to use `.env`

.env.example
```diff
@@ -0,0 +1,29 @@
...
+
+#######################
+# Optional Settings
+#######################
+
+# Cache Directory
+# Location where Curator stores its cache files
+# Default: ~/.curator
+# Note: This can also be set via the working_dir parameter in code
+CURATOR_CACHE_DIR=/path/to/cache/directory
```
CharlieJCJ, line 21: perhaps we can comment out this line or do `CURATOR_CACHE_DIR=`? 

then, in code, check CURATOR_CACHE_DIR exist any non-null value, and use that. 

Currently, you are hard-coding a value that may have overwrite behavior to any defaults in code.

example: https://github.com/ShishirPatil/gorilla/blob/main/berkeley-function-call-leaderboard/.env.example

lavishsaluja: yes fair, made it empty and added handling in prompter.py added the check for null/non-null value

lavishsaluja:
1. updating main.py task -> yes CLI variables are sufficient, should not over engineer this, removed the viewer config variables
2. for now skipping adding this on README as a lot more people will be visiting the repo for documentation of package instead of development so should avoid all noise for them. what do you think?

CharlieJCJ:
> for now skipping adding this on README as a lot more people will be visiting the repo for documentation of package instead of development so should avoid all noise for them. what do you think?

I think should place it somewhere on README to reduce onboarding frictions and have a few words for its purpose can help a lot when we introduce new things that can be configurable.

README.md
```diff
@@ -133,4 +133,20 @@ nvm install 22
 node -v # should print `v22.11.0`
 # verifies the right npm version is in the environment
 npm -v # should print `10.9.0`
+
 ```
```
lavishsaluja, line 137: @CharlieJCJ added this section here - what do you think now? (it is collapsible so that only developers see this and does not add to normal users of the package who come to git to get a quick start guide.

CharlieJCJ: Content is good, for the markdown formatting, I changed a little bit after looking at a few other repos' README. Two nits: 1) we can make the subsection foldable, so user can see an outline of what is under Development. 2) collapse on h4, instead of h2 for arrow aligned with center of the font. 

Try the following

## Development

<details>
<summary><h4 style="display: inline-block">Environment Setup with <code>.env</code> file (optional)</h4></summary>

Copy `.env.example` to create your `.env` file:
```bash
cp .env.example .env
```

Available environment variables:
- `OPENAI_API_KEY`: (Required) Your OpenAI API key for language model requests
- `CURATOR_CACHE_DIR`: (Optional) Custom cache directory, defaults to `~/.cache/curator`. Used to store datasets, model responses, and metadata for caching and viewer functionality.

</details>

lavishsaluja:
@CharlieJCJ just made the change of changing collapsible, thanks for the input. it looks better now.

CharlieJCJ:
Thanks! 

We can merge this after #140 merge, and include part of 0.1.10, since there are some frontend path changes need to address if we introduce the environment. In the meantime, can you resolve merge conflict to dev in `README.md`, thanks

CharlieJCJ:
Another question I'm thinking of is how the README would structure after introducing `.env` because the current workflow is to do pip install, and no where in prompter and viewer do `load_env`. Another thing is if user chooses to do the `.env` thing they would `git clone` the repo first and the prompter and viewer needs to know where this `.env` is, which users would poetry install, etc, but then poetry isn't been mentioned at all in the README. Trying to think what's the best way to do here. cc @madiator

lavishsaluja:
fair @CharlieJCJ - I was just thinking of adding an env for people who want to contribute but I guess best to just totally skip this for now - most people will just use with pip and won't need development side information in README

Issue #145, opened 2024-11-19
[Generic Response Expose Metrics]: `create_by`, `finished_by`


CharlieJCJ:

CharlieJCJ:
+ UI distribution view on duration

CharlieJCJ:
I'm expecting outliers, need to address in distribution views, related #105

Issue #105, opened 2024-11-14
Distribution graph is skewed for data viewer 


vutrung96:
![Screenshot 2024-11-14 at 9 19 36 AM](https://github.com/user-attachments/assets/a04c581d-4c5f-432a-9eb0-a9798c42d069)

It's a bit unhelpful if every almost everything goes into the first bucket. I think this is because we have long tail rows and we split the range (min-max) equally. We should probably exclude outliers  somehow or do log scale.

CharlieJCJ:
Need a combination of https://recharts.org/en-US/examples/BrushBarChart, and adjusted bin size when shifting histogarm ranges.

RyanMarten:
We are using a different graph now.

CharlieJCJ:
Actually, this issue still exist for histogram views that are available in the dropdown.

PR #149, opened 2024-11-19
[curator-viewer] add time logging and curator viewer show distribution

CharlieJCJ:
**Changes**:
- added generation time logging
- add generation time UI, in distribution view, sortable table view and detail view
- added requests & responses overlay time series plot (default view)
- enabled all with streaming
- Resolves 
  - #145 
  - partially address #105 for better bins, but addressing outliers, still need work. Perhaps selecting a range for user to zoom in if they see most of the points lie in a small region?
  - #150

New GenericResponse:
```python
class GenericResponse(BaseModel):
    response_message: Optional[Dict[str, Any]] | str = None
    response_errors: Optional[List[str]] = None
    raw_response: Optional[Dict[str, Any]]
    raw_request: Optional[Dict[str, Any]] = None
    generic_request: GenericRequest
    created_at: datetime.datetime
    finished_at: datetime.datetime
```

https://github.com/user-attachments/assets/a8751cb1-4bf4-4921-b17d-b3fd8726803c

https://github.com/user-attachments/assets/57877738-0d83-49ca-98c7-3ecb8c5a3e7a

Nice to have: 
- https://www.infoworld.com/article/2334935/recharts-intro-to-javascript-charting.html (brush component in UI), for selecting time range

### DIFF
--- a/bespoke-dataset-viewer/app/globals.css
+++ b/bespoke-dataset-viewer/app/globals.css
@@ -37,7 +37,6 @@ body {
     --tooltip-bg: white;
     --tooltip-text-bg: rgb(255, 255, 255);
     --tooltip-text: black;
-    --border: #ccc;
   }
   .dark {
     --background: 0 0% 3.9%;
--- a/bespoke-dataset-viewer/app/globals.css
+++ b/bespoke-dataset-viewer/app/globals.css
@@ -34,6 +34,10 @@ body {
     --chart-5: 27 87% 67%;
     --radius: 0.5rem;
     --success: 142 40% 75%;
+    --tooltip-bg: white;
+    --tooltip-text-bg: black;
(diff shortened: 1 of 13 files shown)


CharlieJCJ:
Ready for review @vutrung96 @RyanMarten

CharlieJCJ:
point1, 2 resolved.
point 3 addressed. 
point 4 is due to not `poetry shell` during testing.

vutrung96 (approved):
can confirm it works on my end and the graph looks beautiful. thank you @CharlieJCJ !

vutrung96 (changes requested):
Screenshot 2024-11-19 at 10 29 45 PM
actually one more issue, seems like text is off here (both lines say responses received)

vutrung96 (approved):

Issue #127, opened 2024-11-15
prompt_func calls another function and that changes, cache does not invalidate


RyanMarten:

RyanMarten:
Related to #148

PR #167, opened 2024-11-21
Use dill pickle to capture the execution context 

vutrung96:
This is what HF relies on to capture the closure variables and global variables

We need to use recurse=True to properly capture changing global variables https://github.com/huggingface/datasets/blob/17f17b3fe7f276e1b019cca8aa651bf7c818a928/src/datasets/utils/_dill.py#L101


CharlieJCJ (approved):
LGTM, suggested nested function tests since this PR also address #127 , thanks @vutrung96 !

Issue #172, opened 2024-11-23
add `clear_cache` for curator


CharlieJCJ:
add method `curator.clear_cache()`

RyanMarten:
right now we just do `rm -rf ~/.cache/curator`

And if you want to disable the cache set

`export CURATOR_DISABLE_CACHE=true`

PR #175, opened 2024-11-26
clear_cache()

lavishsaluja:
for issue - https://github.com/bespokelabsai/curator/issues/172

to test:
1. setup venv
2. set .env with OPENAI_API_KEY
3. install the curator package using local directory and not PyPi package
4. run `python test_clear_cache_integration.py` under tests to generate a dataset -> clear it using new clear_cache function
5. use it in collab

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .venv
 __pycache__
 .vscode
+.env
 
 # Static / Build
 /src/bespokelabs/curator/viewer/static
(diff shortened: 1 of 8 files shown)


CharlieJCJ:
Comments: 
- [ ] resolve merge conflicts
- [ ] lints proposed by devin that I don't think should be part of  in https://github.com/bespokelabsai/curator/pull/175/files#diff-aed42848867b17c3aaad8db31996378f344ef6d28d354ae7f82e01c13a3990ec

PR #181, opened 2024-11-30
Allow special tokens when encoding text for token accounting

### LINKED ISSUE #178: Curator Fails on Disallowed Special Token <|endoftext|>
The Curator library encounters a failure during token encoding due to the <|endoftext|> token being disallowed when a text contains the <|endoftext|> token. The error occurs when processing requests with the curator/request_processor/openai_online_request_processor.py.

The relevant error message is: 
raise ValueError(
ValueError: Encountered text corresponding to disallowed special token '<|endoftext|>'.
If you want this text to be encoded as a special token, pass it to `allowed_special`, e.g. `allowed_special={'<|endoftext|>', ...}`.
If you want this text to be encoded as normal text, disable the check for this token by passing `disallowed_special=(enc.special_tokens_set - {'<|endoftext|>'})`.
To disable this check for all special tokens, pass `disallowed_special=()`.

This can be fixed by disabling the check or resorting to encoding the token as special token or as normal text.

RyanMarten:
Fixes #178

### DIFF
--- a/src/bespokelabs/curator/request_processor/openai_online_request_processor.py
+++ b/src/bespokelabs/curator/request_processor/openai_online_request_processor.py
@@ -674,7 +674,7 @@ def num_tokens_consumed_from_request(
                 num_tokens += 4  # every message follows <im_start>{role/name}\n{content}<im_end>\n
                 for key, value in message.items():
                     try:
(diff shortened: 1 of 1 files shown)


vutrung96 (approved):
Tested yet?

PR #200, opened 2024-12-04
Increase max retry to 50

vutrung96:
We are still hitting timeouts from OpenAI. Will retry much more aggressively.

```
Error Type: TASK_EXECUTION_EXCEPTION

User exception:
10_067307_1/runtime_resources/pip/14f34acc790047a95e39a14509b14c99a0377d40/virtualenv/lib/python3.10/site-packages/openai/_base_client.py", line 1581, in _request
 return await self._retry_request(
 File "/tmp/ray/session_2024-11-27_21-23-10_067307_1/runtime_resources/pip/14f34acc790047a95e39a14509b14c99a0377d40/virtualenv/lib/python3.10/site-packages/openai/_base_client.py", line 1591, in _request
 raise APITimeoutError(request=request) from err
openai.APITimeoutError: Request timed out.
```

Note that OpenAI client will apply exponential backoff, up to at most 8 seconds: https://github.com/openai/openai-python/blob/bb9c2de913279acc89e79f6154173a422f31de45/src/openai/_base_client.py#L704, so we are not bombarding OpenAI's server with requests here.

### DIFF
--- a/src/bespokelabs/curator/request_processor/openai_batch_request_processor.py
+++ b/src/bespokelabs/curator/request_processor/openai_batch_request_processor.py
@@ -1,5 +1,6 @@
 import asyncio
 import datetime
+import glob
 import json
 import logging
 import os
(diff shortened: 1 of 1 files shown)


CharlieJCJ (approved):
LGTM

Issue #226, opened 2024-12-07
Resubmit expired requests in expired batches, while retaining successful responses


RyanMarten:
Should at least use the responses that have been completed. 

Ofc the best thing would be to resubmit only requests that weren't completed. 

Instead of resubmitting the whole batch that expired. 

But the issue with this is now there isn't a 1 to 1 connection between requests --> batch --> responses. Keeping it simple we can just submit a smaller batch which is the remaining amount for that original batch.

RyanMarten:
Right now we are not downloading the output or error files:
https://github.com/bespokelabsai/curator/blob/dev/src/bespokelabs/curator/request_processor/openai_batch_request_processor.py#L946-L977

RyanMarten:
And does not create a response fie: 
https://github.com/bespokelabsai/curator/blob/dev/src/bespokelabs/curator/request_processor/openai_batch_request_processor.py#L1004-L1007

RyanMarten:
The first thing we should do just allow for brute-force retry (aka run the same program again, Curator treats it as a completely failed batch and resubmits). 

Then we should add the more fine-grained solution

RyanMarten:
On (1) all we need to do is add an if statement when we check which batches are already submitted and don't mark as submitted if failed or expired or cancelled 

https://github.com/bespokelabsai/curator/blob/dev/src/bespokelabs/curator/request_processor/openai_batch_request_processor.py#L721-L755

RyanMarten:
We also want to resubmit requests that don't have a valid response format #86 

(although very expensive to do if we resubmit the whole batch multiple times, so just submit failed requests)

RyanMarten:
Responses in the file will look like this

https://platform.openai.com/docs/guides/batch/getting-started?lang=node#batch-expiration
```
{"id": "batch_req_123", "custom_id": "request-3", "response": null, "error": {"code": "batch_expired", "message": "This request could not be executed before the completion window expired."}}
{"id": "batch_req_123", "custom_id": "request-7", "response": null, "error": {"code": "batch_expired", "message": "This request could not be executed before the completion window expired."}}
```

RyanMarten:
When we get failed requests (or requests with finish_reason of length or content filter, resubmit these in batch). 
for the finish_reason we can use the litellm map. Started PR for improving that function.  
- https://github.com/BerriAI/litellm/pull/7264

PR #231, opened 2024-12-09
Various fixes to increase the reliability of batch processing

vutrung96:
* Recover using submitted batches from other API keys --- if they are from a different project and the batches are not found we simply move on
* Retry more aggressively in OpenAI client
* Handle failures more gracefully instead of error-ing out in various places

### DIFF
--- a/src/bespokelabs/curator/request_processor/openai_batch_request_processor.py
+++ b/src/bespokelabs/curator/request_processor/openai_batch_request_processor.py
@@ -1014,7 +1014,9 @@ class BatchManager:
                     logger.warning(f"Failed to delete file {file_id}")
             except NotFoundError:
                 # This is fine, the file may have been deleted already. Deletion should be best-effort.
-                logger.warning(f"Trying to delete file {file_id} but it was not found, skipping deletion.")
+                logger.warning(
+                    f"Trying to delete file {file_id} but it was not found, skipping deletion."
+                )
 
     async def download_batch(self, batch: Batch) -> str | None:
         file_content = None
--- a/src/bespokelabs/curator/request_processor/openai_batch_request_processor.py
+++ b/src/bespokelabs/curator/request_processor/openai_batch_request_processor.py
@@ -516,7 +516,7 @@ class BatchManager:
             delete_failed_batch_files (bool): Whether to delete input/error files from OpenAI
                 after batch failure.
         """
-        self.client = AsyncOpenAI()
+        self.client = AsyncOpenAI(max_retries=MAX_RETRIES_PER_OPERATION)
         self.check_interval = check_interval
         self.working_dir = working_dir
         self.tracker = BatchStatusTracker()
--- a/src/bespokelabs/curator/request_processor/openai_batch_request_processor.py
+++ b/src/bespokelabs/curator/request_processor/openai_batch_request_processor.py
@@ -784,7 +784,9 @@ class BatchManager:
                 self.tracker.mark_as_submitted(request_file_name, batch_object, n_requests)
             else:
                 # batch objects if not unsubmitted, should be downloaded
-                assert batch_object.id in self.tracker.downloaded_batches
+                raise ValueError(
+                    f"Batch {batch_object.id} is not unsubmitted, but not in tracker.downloaded_batches: {self.tracker.downloaded_batches}"
+                )
 
         if self.tracker.n_submitted_batches > 0:
             logger.info(
@@ -802,6 +804,9 @@ class BatchManager:
             glob.glob(f"{self.working_dir}/batch_objects_downloaded_*.jsonl")
         )
         for downloaded_batch_object_file in downloaded_batch_object_files:
+            logger.info(
+                f"Processing downloaded batch objects file: {downloaded_batch_object_file} Your API key is ***{self.client.api_key[-4:]}."
+            )
             with open(downloaded_batch_object_file, "r") as f:
                 for line in f:
                     batch_object = Batch.model_validate(json.loads(line))
--- a/src/bespokelabs/curator/request_processor/openai_batch_request_processor.py
+++ b/src/bespokelabs/curator/request_processor/openai_batch_request_processor.py
@@ -780,18 +780,18 @@ class BatchManager:
                     ):
                         existing_submitted_batches[request_file_name] = batch_object
 
-        logger.info(f"existing_submitted_batches: {existing_submitted_batches}")
         for request_file_name, batch_object in existing_submitted_batches.items():
 
             output_file_id = batch_object.output_file_id
-            try:
-                await self.client.files.retrieve(output_file_id)
-            except NotFoundError:
-                logger.warning(
-                    f"Output file {output_file_id} not found for batch {batch_object.id}. "
-                    "The file may have been deleted. Will resubmit this batch..."
-                )
-                continue
+            if output_file_id is not None:
+                try:
+                    await self.client.files.retrieve(output_file_id)
+                except NotFoundError:
+                    logger.warning(
+                        f"Output file {output_file_id} exists in batch object but cannot be found "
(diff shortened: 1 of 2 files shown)


src/bespokelabs/curator/request_processor/openai_batch_request_processor.py
```diff
@@ -721,20 +719,51 @@ async def submit_batch_from_request_file(
...
                     logger.debug(
                         f"Already submitted batch {batch_object.id} for request file {request_file_name}. "
                         f"Getting batch object to update tracker."
                     )
-                    batch_object = await self.retrieve_batch(batch_object.id)
+                    try:
+                        batch_object = await self.retrieve_batch(batch_object.id)
+                    except NotFoundError:
+                        logger.info(
+                            f"Batch {batch_object.id} not found. This is fine since we might be looking at a cancelled batch, "
```
RyanMarten, line 749: Can we make this log more self-contained? 
E.g. Already submitted batch {} not found and will be resubmitted. This can happen if the batch is submitted within a different project. 

(There is no API for deleting a batch. Also cancelled batches return an object from retrieval API). 
Also warning?

src/bespokelabs/curator/request_processor/openai_batch_request_processor.py
```diff
@@ -721,20 +719,51 @@ async def submit_batch_from_request_file(
...
+                            f"a deleted batch, or a batch from another project. Will continue..."
+                        )
+                        continue
+
+                    if not self._validate_batch_status(batch_object.status):
+                        logger.info(
```
RyanMarten, line 755: again warning.
and again, say that this batch was already submitted and will be resubmitted

src/bespokelabs/curator/request_processor/openai_batch_request_processor.py
```diff
@@ -937,11 +1006,17 @@ async def delete_file(self, file_id: str, semaphore: asyncio.Semaphore):
...
+            try:
+                delete_response = await self.client.files.delete(file_id)
+                if delete_response.deleted:
+                    logger.debug(f"Deleted file {file_id}")
+                else:
+                    logger.warning(f"Failed to delete file {file_id}")
+            except NotFoundError:
+                # This is fine, the file may have been deleted already. Deletion should be best-effort.
+                logger.warning(
+                    f"Trying to delete file {file_id} but it was not found, skipping deletion."
```
RyanMarten, line 1018: no need to say "skipping deletion".

src/bespokelabs/curator/request_processor/openai_batch_request_processor.py
```diff
@@ -743,18 +772,49 @@ async def track_already_submitted_batches(self):
...
+                        f"tracker.downloaded_batches: {self.tracker.downloaded_batches} \n"
+                    )
 
         if self.tracker.n_submitted_batches > 0:
             logger.info(
                 f"{self.tracker.n_submitted_batches:,} out of {self.tracker.n_total_batches - self.tracker.n_downloaded_batches:,} remaining batches are already submitted."
             )
 
-    def track_already_downloaded_batches(self):
+    async def track_already_downloaded_batches(self):
```
RyanMarten, line 817: what part of this is async? I believe we removed this before?

src/bespokelabs/curator/request_processor/openai_batch_request_processor.py
```diff
@@ -765,13 +825,23 @@ def track_already_downloaded_batches(self):
...
                     request_file = batch_object.metadata["request_file_name"]
                     response_file = request_file_to_response_file(request_file, self.working_dir)
-                    assert request_file in self.tracker.unsubmitted_request_files
-                    assert os.path.exists(response_file)
+                    assert (
+                        request_file in self.tracker.unsubmitted_request_files
+                    ), f"request_file {request_file} not in unsubmitted_request_files: {self.tracker.unsubmitted_request_files}"
+                    if not os.path.exists(response_file):
+                        logger.warning(
+                            f"response_file {response_file} does not exist. will continue..."
```
RyanMarten, line 841: Again, say that this file doesn't exist during tracking of downloaded batch files and will resubmit.

src/bespokelabs/curator/request_processor/openai_batch_request_processor.py
```diff
@@ -743,18 +772,49 @@ async def track_already_submitted_batches(self):
...
                         n_requests = batch_object.request_counts.total
 
-                    if request_file_name in self.tracker.unsubmitted_request_files:
-                        self.tracker.mark_as_submitted(request_file_name, batch_object, n_requests)
-                    else:
-                        # batch objects if not unsubmitted, should be downloaded
-                        assert batch_object.id in self.tracker.downloaded_batches
+                    # For each request file, we only want to keep the latest batch object.
+                    if (
+                        request_file_name not in existing_submitted_batches
```
RyanMarten, line 777: When would there be multiple valid (found, not expired/cancelled) submitted batches?

src/bespokelabs/curator/request_processor/openai_batch_request_processor.py
```diff
@@ -743,18 +772,49 @@ async def track_already_submitted_batches(self):
...
+                        or existing_submitted_batches[request_file_name].created_at
+                        < batch_object.created_at
+                    ):
+                        existing_submitted_batches[request_file_name] = batch_object
+
+        for request_file_name, batch_object in existing_submitted_batches.items():
+
+            output_file_id = batch_object.output_file_id
+            if output_file_id is not None:
+                try:
```
RyanMarten, line 787: If the batch failed, the it has an error_file_id instead. 
Also with expired batches we have both output and error file I believe. 
But we can ignore that complexity. And just go with what you have here (so if it failed, we resubmit, and if it expired we aren't dealing with it right now). 
Although we should have a fix that resubmits a smaller batch only containing failed requests and then combine together later as discussed in #226

RyanMarten (approved):
Left a couple of comments

src/bespokelabs/curator/request_processor/openai_batch_request_processor.py
vutrung96: Fixed.

src/bespokelabs/curator/request_processor/openai_batch_request_processor.py
vutrung96: It might or might not be resubmitted as I explained above :)

src/bespokelabs/curator/request_processor/openai_batch_request_processor.py
vutrung96: This is very unlikely if we just let curator do its thing but can happen if we are doing some sketchy cache surgery stuff. I'm inclined to keep it just in case we fall into some weird state like that.

src/bespokelabs/curator/request_processor/openai_batch_request_processor.py
vutrung96: If a batch object either expired or failed, it is not added existing_submitted_batches, so it will not be considered here and we will resubmit that batch.

I'm inclined to make each batch "atomic" and either fail or pass the whole batch. These batches are small enough (10k) that I think we should just resubmit everything on failure instead of trying to recover. That will simplify our logic.

src/bespokelabs/curator/request_processor/openai_batch_request_processor.py
vutrung96: ooops sorry, don't know what added it back for some reason.

src/bespokelabs/curator/request_processor/openai_batch_request_processor.py
vutrung96: Done.

PR #247, opened 2024-12-12
Add more docstrings and remove checks which are already in PromptFomatter

madiator:
1. Docstrings not present for new args added. 
2. The types for the args are well defined in __init__(), so no need to repeat them in docstrings.
3. The error checking is handled in PromptFormatter, so no need to check again.


RyanMarten (approved):
LGTM

PR #248, opened 2024-12-12
More graceful handling, pt 2

vutrung96:


src/bespokelabs/curator/request_processor/base_request_processor.py
```diff
@@ -236,7 +236,16 @@ def create_dataset_files(
...
                     for generic_response_string in f_in:
                         total_responses_count += 1
-                        response = GenericResponse.model_validate_json(generic_response_string)
+
+                        try:
+                            response = GenericResponse.model_validate_json(generic_response_string)
+                        except ValidationError as e:
+                            logger.warning(
+                                f"Failed to parse response message as JSON: {generic_response_string}. "
+                                f"The model likely returned an invalid JSON format. Will skip this response."
```
madiator, line 245: this doesn't need to be a f-string.

src/bespokelabs/curator/request_processor/base_request_processor.py
```diff
@@ -236,7 +236,16 @@ def create_dataset_files(
...
+                            )
+                            failed_responses_count += 1
+                            continue
```
madiator, line 248: These continue statements are not fun. Make the code a bit non-linear to read. 
A lot of the logic should be under try itself.

PR #254, opened 2024-12-13
Revert "Refactor LLM class to use context manager for batch parameters"

madiator:
Reverts bespokelabsai/curator#250.
The semantics is not yet good.


RyanMarten (approved):
reverting

PR #255, opened 2024-12-14
Add a SimpleLLM interface, and update documentation.

madiator:


.DS_Store
RyanMarten, line 1: can you delete this file please?

docs/curator-dataset.png
RyanMarten, line 1: can you choose a more interesting screenshot than 1 request and 1 response?

README.md
```diff
@@ -38,43 +38,83 @@ Bespoke Curator is an open-source project:
...
 class Poems(BaseModel):
     poems_list: List[Poem] = Field(description="A list of poems.")
+```
 
-
-# We define an `LLM` object that generates poems which gets applied to the topics dataset.
+We define an `LLM` object that generates poems which gets applied to the topics dataset.
+```python
 poet = curator.LLM(
     # `prompt_func` takes a row of the dataset as input.
```
RyanMarten, line 119: I find the comments here busy and confusing. 
put all before in a """ comment?

src/bespokelabs/curator/llm/llm.py
```diff
@@ -37,79 +37,6 @@
 class LLM:
     """Interface for prompting LLMs."""
```
RyanMarten, line 39: I think that __init__ should be the first function in the class?

RyanMarten (approved):
some small comments

src/bespokelabs/curator/llm/llm.py
madiator: Yes, absolutely! I was planning to mention this and move it, but for clarity thought i will remove the incorrect one and keep the correct one and then move it. Moved it now.

README.md
madiator: How about now?

madiator (commented):
addressed

docs/curator-dataset.png
madiator: done

PR #257, opened 2024-12-14
Add metadata dict + cache verification

GeorgiosSmyrnis:
This PR adds a metadata dict to the request file generation - this should allow for better handling of request files + checking the integrity of the cache.

### DIFF
--- a/src/bespokelabs/curator/request_processor/base_request_processor.py
+++ b/src/bespokelabs/curator/request_processor/base_request_processor.py
@@ -221,12 +221,12 @@ class BaseRequestProcessor(ABC):
                 request = prompt_formatter.create_generic_request(dataset_row, dataset_row_idx)
                 await f.write(json.dumps(request.model_dump(), default=str) + "\n")
 
-        num_written = end_idx - start_idx
-        metadata_dict = {"num_jobs": num_written}
+        num_requests = end_idx - start_idx
+        metadata_dict = {"num_jobs": num_requests}
         async with aiofiles.open(metadata_file, "w") as f:
             await f.write(json.dumps(metadata_dict, indent=4) + "\n")
 
-        logger.info(f"Wrote {num_written} requests to {request_file}.")
+        logger.info(f"Wrote {num_requests} requests to {request_file}.")
 
     def attempt_loading_cached_dataset(
         self, working_dir: str, parse_func_hash: str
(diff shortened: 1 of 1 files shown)


src/bespokelabs/curator/request_processor/base_request_processor.py
```diff
@@ -168,7 +220,13 @@ async def acreate_request_file(
                 # Get the generic request from the map function
                 request = prompt_formatter.create_generic_request(dataset_row, dataset_row_idx)
                 await f.write(json.dumps(request.model_dump(), default=str) + "\n")
-        logger.info(f"Wrote {end_idx - start_idx} requests to {request_file}.")
+
+        num_written = end_idx - start_idx
+        metadata_dict = {"num_jobs": num_written}
```
RyanMarten, line 225: nit: num_requests 

also... instead of `metadata_i.json` we can also just store a single `metadata.json` that maps from request file name to metadata object. Just an option to consider. 

Then we can also store another field which is the total number of request files (which also helps with validation) E.g. if in the middle of batch creation the worker dies and only some of the request files are present we wouldn't catch that if we just looked at each file.

RyanMarten (changes requested):
Feedback in comments

src/bespokelabs/curator/request_processor/base_request_processor.py
GeorgiosSmyrnis: @RyanMarten thanks for the comment - having single metadata json would require some changes to the structure (since currently these async functions don't really return anything).

I changed it to still having `metadata_i.json`, but instead of only verifying if all exist it also explicitly checks each file separately, and regenerating only missing ones. This takes care of the second case you mention as well.

GeorgiosSmyrnis:
This should be ready now - please have a look

src/bespokelabs/curator/request_processor/base_request_processor.py
```diff
@@ -76,6 +77,51 @@ def run(
         """
         pass
 
+    def _get_bad_cache_files(self, working_dir: str, dataset: Optional[Dataset]) -> List[int]:
```
RyanMarten, line 80: small nit: 

can we rename this to _verify_existing_request_files

or something that is specific to request files (because there are a bunch of other types of files in the cache) and it is unclear what `bad` is.

RyanMarten (approved):
LGTM!

PR #262, opened 2024-12-15
Raise on None response message

RyanMarten:


src/bespokelabs/curator/request_processor/base_online_request_processor.py
```diff
@@ -275,6 +275,11 @@ async def process_requests_from_file(
                                 f"{response.response_errors}, removing from output and will retry"
                             )
                             num_previously_failed_requests += 1
+                        if response.response_message is None:
+                            logger.debug(
```
madiator, line 279: why debug? maybe warning?

madiator (approved):

src/bespokelabs/curator/request_processor/base_online_request_processor.py
RyanMarten: this is when resuming from previously started run. 

At the end of reading all the responses it will info log the total failed requests. 
This way it doesn't spam with 1,000s of warning logs

Issue #261, opened 2024-12-15
Detect finish_reason that is not `stop`


RyanMarten:
Gemini returns None messages when content_filter is the finish reason. 

Other finish reasons like (maximum length) should be detected and dealt with. The default behavior should not be ignoring.

RyanMarten:
Also related, retry logic based on the error encountered. 

E.g. do we want to retry on content_filter? On hitting max tokens instead of stop token, etc.
- #86

RyanMarten:
litellm can access like this: 
https://docs.litellm.ai/docs/completion/output

RyanMarten:
litellm provides a map between all the finish reasons to standard openai set

https://github.com/BerriAI/litellm/blob/fd583e715ec1b183cade08ac29c1cbe73e5c94f6/litellm/litellm_core_utils/core_helpers.py#L18-L54

Do they process all output with this?

RyanMarten:
So what was actually going on is RECITATION which litellm maps to content filter (in the function I linked above).

RyanMarten:
Short term, I don't think there is actually any solution to turn off the RECITATION content filter. 

Others are also seeing this filter randomly on prompts that seem innocuous: 
- https://github.com/BerriAI/litellm/issues/2866
- https://github.com/GoogleCloudPlatform/generative-ai/issues/344#issuecomment-2014706679

They claim to have fixed this: 
- https://github.com/GoogleCloudPlatform/generative-ai/issues/344#issuecomment-2284117249
> It looks like this issue was primarily caused by the language preprocessor classifying certain prompts (especially those with many numbers or mixed languages) as unsupported, leading to the "recitation blocking"

But I'm still seeing the recitation blocking behave poorly.

This issue lists out some other discussions / resources 
- https://github.com/gbaptista/gemini-ai/issues/21

The official docs currently say 
https://cloud.google.com/vertex-ai/docs/reference/rest/v1beta1/GenerateContentResponse#FinishReason
> RECITATION	The token generation stopped because of potential recitation.

More discussion: 
- https://github.com/google/generative-ai-docs/issues/257

People are suggestion asking for > 1 completions for every request, but that is unnecessarily expensive.

RyanMarten:
It might be a good idea to add this as part of the `generic_response` object and have both litellm and openai write to that and handle this in the base request processor

RyanMarten:
OpenAI finish reason options: 
[...]

PR #266, opened 2024-12-16
Retry on response format failure

### LINKED ISSUE #86: Retry when structured output fails
Thing things we need to check and then retry if they fail
(1) the response is valid JSON (try-catch current in `parse_response_message`) 
(2) the response JSON matches the Pydantic model schema (base_request_processor in `create_dataset_files`) 

Right now these checks happen after the LLM responds (successfully sends back a string) so we don't have it within the retry logic. For batch, we would need to accumulate all the failed responses and send a new batch. 
For Online, we can add these checks before writing to `responses.jsonl` file. However the logic might be weird since everything about this parsing was abstracted away from the requests --> responses logic in online. Might need to rethink the pathways a little bit. 

Lessons learned from #85

RyanMarten:
Fixes #86

### DIFF
--- a/src/bespokelabs/curator/request_processor/base_request_processor.py
+++ b/src/bespokelabs/curator/request_processor/base_request_processor.py
@@ -262,6 +262,58 @@ class BaseRequestProcessor(ABC):
                     "Deleted file and attempting to regenerate dataset from cached LLM responses."
                 )
 
+    def convert_response_to_response_format(
+        self, response_message: str | dict, response_format: Optional[BaseModel]
+    ) -> Optional[dict | str]:
+        """
+        Converts a response message to a specified Pydantic model format.
+
+        This method takes a response message (either as a string or dict) and validates/converts it
+        according to the provided Pydantic model format. If the response message is a string,
+        it first attempts to parse it as JSON. The resulting dict is then used to construct
+        an instance of the specified Pydantic model.
+
+        Args:
+            response_message (str | dict): The response message to convert, either as a JSON string
+                or a dictionary.
+            response_format (Optional[BaseModel]): The Pydantic model class that defines the
+                expected format of the response.
+
+        Returns:
+            Optional[dict | str]: The validated response message as a Pydantic model instance.
+
+        Raises:
+            json.JSONDecodeError: If the response_message is a string but cannot be parsed as valid JSON.
+            ValidationError: If the parsed response does not match the schema defined by response_format.
+        """
+        # Response message is a string, which is converted to a dict
+        # The dict is then used to construct the response_format Pydantic model
+        try:
+            # First try to parse the response message as JSON
+            if isinstance(response_message, str):
+                try:
+                    response_dict = json.loads(response_message)
+                except json.JSONDecodeError as e:
+                    logger.warning(
+                        f"Failed to parse response message as JSON: {response_message}. "
+                        f"The model likely returned an invalid JSON format."
+                    )
+                    raise e
+            else:
+                response_dict = response_message
+
+            # Then construct the Pydantic model from the parsed dict
+            response_message = response_format(**response_dict)
+            return response_message
+
+        except ValidationError as e:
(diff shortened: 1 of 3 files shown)


RyanMarten:
As discussed in #86, using `strict: true` could help. This argument make sense to expose through the context manager that Mahesh was working on. 
It only an available argument for some models (subset of openai models). Should we verify ourselves with nice errors about supported param? Litellm does that. Or do we have a more sophisticated retry logic (based off tenacity - as Mahesh suggested and noted in #261 ) that doesn't retry when the response code says the request shape is not value.

RyanMarten:
Looking for examples for a minimal reproduction. @neginraoof can maybe provide. 

In https://rwilinski.ai/posts/benchmarking-llms-for-structured-json-generation/ the gpt-4o-mini gets 100% on all the attempts.

RyanMarten:
Had claude help come up with an example


Which the model struggles at creating a valid response to

```
from bespokelabs.curator import LLM
from datasets import Dataset
from pydantic import BaseModel, Field
from typing import List, Dict, Optional
from enum import Enum
from datetime import datetime

# Complex dataset with tricky prompts
dataset = Dataset.from_dict({
    "prompt": [
        """Analyze this fictional company's performance and provide a detailed breakdown:
        TechCorp Industries faced challenges in Q3 2023. Their cloud division struggled 
        while IoT solutions showed promise. Customer satisfaction varied by region."""
    ]
})

class SentimentLevel(str, Enum):
    VERY_NEGATIVE = "very_negative"
    NEGATIVE = "negative"
    NEUTRAL = "neutral"
    POSITIVE = "positive"
    VERY_POSITIVE = "very_positive"

class MetricScore(BaseModel):
    value: float
    confidence_interval: List[float] = Field(min_items=2, max_items=2)
    reliability_score: float = Field(ge=0, le=1)
    last_updated: datetime

class DepartmentAnalysis(BaseModel):
    department_name: str
    revenue_change: float
    key_metrics: Dict[str, MetricScore]
    risk_factors: List[Dict[str, float]]
    sentiment: SentimentLevel
    regional_breakdown: Dict[str, Dict[str, float]]
    future_projections: List[Dict[str, Optional[float]]]

class CompanyAnalysis(BaseModel):
    company_overview: Dict[str, str]
    financial_health: float = Field(ge=-1, le=1)
    departments: List[DepartmentAnalysis]
    market_position: Dict[str, List[float]]
    competitor_comparison: Dict[str, Dict[str, float]]
    risk_assessment: List[Dict[str, Optional[float]]]
    timestamp: datetime
    confidence_score: float = Field(ge=0, le=1)

prompter = LLM(
    prompt_func=lambda row: row["prompt"],
    model_name="gpt-4o-mini",
    response_format=CompanyAnalysis,
)

dataset = prompter(dataset)
print(dataset.to_pandas())
```

src/bespokelabs/curator/request_processor/base_request_processor.py
```diff
@@ -262,6 +262,58 @@ def attempt_loading_cached_dataset(
                     "Deleted file and attempting to regenerate dataset from cached LLM responses."
                 )
 
+    def convert_response_to_response_format(
+        self, response_message: str | dict, response_format: Optional[BaseModel]
+    ) -> Optional[dict | str]:
+        """
+        Converts a response message to a specified Pydantic model format.
```
madiator, line 269: A bit lost on the motivation here.

RyanMarten: this function we use anyways, I just moved it out so we don't duplicate code

We need to convert to the correct response format from the string that model returns

madiator: Ah sorry.
Also, does it make sense to put this in prompt_formatter?
So `self.prompt_formatter.convert_response_to_response_format(response_message)`.

madiator (approved):

PR #274, opened 2024-12-18
Clean ups in llm.py

madiator:
Fix minor issues. For example, we doin't need a single line __call__ method to call _completions method.


RyanMarten (approved):
LGTM

PR #275, opened 2024-12-18
Put the examples in respective folders and add requirements.txt everywhere

madiator:
Also rename from "camel.py" which not many people understand to "ungrounded-qa"


RyanMarten (approved):
LGTM 

Tested the examples. 

It might make sense to have a README in each folder, or at least in the high level /examples folder.

madiator:
I added a high level README. We can add per folder if that makes more sense.

Issue #288, opened 2024-12-24
Ordering should be preserved between input and output.


madiator:
The output ordering is different each time. Here is an example where the inputs are ordered, but after execution the outputs are in different order.
<img width="487" alt="image" src="https://github.com/user-attachments/assets/99425ec9-e5fc-4a6d-8fd6-eaa479ed41f6" />
This is not desirable.

Issue #276, opened 2024-12-19
User messages and assistant messages are not mapped correctly


madiator:
User messages and assistant messages need to be mapped correctly. So sometimes, the user messages get duplicated.
![image](https://github.com/user-attachments/assets/59c57e38-c630-4d1a-83b9-e1eaa6ebae9b)

madiator:
Update: it looks like this is not an issue with the viewer, but affecting all of batch generation.

madiator:
Something off with the processing of batches.

```
{'input': 'What is the value of 3*3', 'answer': 4}
{'input': 'What is the value of 3*3', 'answer': 9}
{'input': 'What is the value of 1*1', 'answer': 0}
{'input': 'What is the value of 1*1', 'answer': 1}
{'input': 'What is the value of 5*5', 'answer': 16}
{'input': 'What is the value of 5*5', 'answer': 25}
```

PR #289, opened 2024-12-25
Fix bug in batch mapping and get right order for outputs.

madiator:
1. Bug in batch mapping: https://github.com/bespokelabsai/curator/issues/276
We just need to map the response to the right generic request. This is a simple fix (but the bug costed a lot of dollars).
   
2. Ordering wasn't being respected earlier: https://github.com/bespokelabsai/curator/issues/288
We need to keep track of the original row idx and sort by it (and then drop that column).
Added a test that tests both the above.

3. Also minor change: replace the deprecated schema_json() with model_json_schema().

### DIFF
--- a/src/bespokelabs/curator/request_processor/base_request_processor.py
+++ b/src/bespokelabs/curator/request_processor/base_request_processor.py
@@ -365,7 +365,8 @@ class BaseRequestProcessor(ABC):
                                 raise ValueError(
                                     f"Got empty row {row} from `parse_func`. {error_help}"
                                 )
-
+                            # Add the original row index to the row so that we can sort by it later.
+                            row["__original_row_idx"] = response.generic_request.original_row_idx
                             writer.write(row)
 
             logger.info("Finalizing writer")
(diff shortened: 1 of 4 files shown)


src/bespokelabs/curator/request_processor/base_request_processor.py
```diff
@@ -398,7 +399,10 @@ def create_dataset_files(
                         f"Some requests do not have responses and require_all_responses is True."
                     )
 
-        return Dataset.from_file(dataset_file)
+        d = Dataset.from_file(dataset_file)
```
RyanMarten, line 402: Why do we need to sort?

RyanMarten (approved):
For the fix, I don't believe we need to sort. 
This may add extra overhead, but is better for the end-user.

src/bespokelabs/curator/request_processor/base_request_processor.py
madiator: If I sent requests [req1, req2, req3] (say through a dataset), the ordering of the responses can be anything right now. And varies with each run.
It's a bit confusing, and most people expect the responses to be in the order they sent the queries.

madiator:
The overhead is very little, compared to the network costs. Sorting 10k toy examples takes only 25ms for me.

PR #295, opened 2024-12-29
Retry failed requests within batches

### LINKED ISSUE #226: Resubmit expired requests in expired batches, while retaining successful responses

### LINKED ISSUE #296: Resubmit failed requests in batch mode
For example, a number of batches were submit to anthropic. Because of low funds in the account, hundreds or thousands of requests in each batch failed. 

Screenshot 2024-12-29 at 3 42 13 PM

The lines in the batch file look like this: 
```
{"custom_id":"999974","result":{"type":"errored","error":{"type":"error","error":{"details":null,"type":"invalid_request_error","message":"Your credit balance is too low to access the Anthropic API. Please go to Plans & Billing to upgrade or purchase credits."}}}}
```

Related to #226. Both OpenAI and Anthropic have similar behavior, so let's implement this in the base batch request processor in a new branch off of #243.

RyanMarten:
Retrying a request within a batch

closes #296
closes #226 

also address #86 in batch mode


tests/batch/test_partial_batch_resubmit.py
```diff
@@ -0,0 +1,116 @@
+import json
```
adamoptimizer, line 1: Lets move this to test_all

adamoptimizer (changes requested):

src/bespokelabs/curator/request_processor/base_request_processor.py
```diff
@@ -427,3 +427,72 @@ def _load_from_dataset_file(self, dataset_file: str) -> Dataset:
...
+
+        if not os.path.exists(metadata_file):
+            raise ValueError(f"Metadata file not found: {metadata_file}")
+
+        try:
+            with open(metadata_file, "r") as f:
+                metadata = json.load(f)
+                return metadata
+        except json.JSONDecodeError as e:
+            raise ValueError(f"Invalid JSON in metadata file: {metadata_file}") from e
```
vutrung96, line 498: In this case we should probably ask users to delete cache?

vutrung96, line 498: In this case we should probably ask users to delete cache?

src/bespokelabs/curator/request_processor/batch/base_batch_request_processor.py
```diff
@@ -444,22 +447,33 @@ async def submit_batches_from_request_files(
...
-            n_remaining = self.tracker.n_total_batches - self.tracker.n_downloaded_batches
-            n_submitted = self.tracker.n_submitted_batches + self.tracker.n_finished_batches
-            logger.info(f"{n_submitted:,} out of {n_remaining:,} remaining batches previously submitted.")
-        # exit early
-        if self.tracker.n_unsubmitted_request_files == 0:
+
+        tasks = []
+
+        # check existing response files for resuming
+        for batch in self.tracker.downloaded_batches.values():
```
vutrung96, line 454: don't we also need to do this retry logic in poll_and_process_batch too? otherwise we're only retrying failed requests across restarts, but not during a job.

or do you basically say, if you set requires_all_responses to True, then your job will fail and then you can restart?

the reason this manual failing and retrying can be annoying is if we run batch completions multiple times in a pipeline and it keeps failing and we have to manually resume.

vutrung96 (changes requested):

src/bespokelabs/curator/request_processor/base_request_processor.py
adamoptimizer: agreed

src/bespokelabs/curator/request_processor/batch/base_batch_request_processor.py
adamoptimizer: @vutrung96 we discussed this yesterday,
I guess this could be handled separately in a PR,

 I was thinking to add an `retry_queue`, polling function could populate this `retry_queue` with failed requests.
and we can batch them for retry 
just thinking out load on this, not fixated to something.

adamoptimizer (approved):
Thanks for the pr!

tests/conftest.py
```diff
@@ -31,6 +38,7 @@ def temp_working_dir(request):
         yield temp_working_dir, backend, vcr_config
     finally:
         shutil.rmtree(temp_working_dir)
+        pass
```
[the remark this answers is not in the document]
adamoptimizer: u can remove before merging, I approved

src/bespokelabs/curator/request_processor/base_request_processor.py
RyanMarten: Did we add .clear_cache back or should I just say `rm -rf parent(metadata_file)`

src/bespokelabs/curator/request_processor/batch/base_batch_request_processor.py
RyanMarten: > or do you basically say, if you set requires_all_responses to True, then your job will fail and then you can restart?

Yes exactly. 

For now this actually worked well in my experience testing it with sonnet.  

I also think this is not a bad idea since some requests will always fail (e.g. content_filter or malformed request in some way) so we could get into an infinite loop of submitting batches. We could only do the automatic retries for certain failures (expiration - but that only happens after 24 hours anyways), but let's make that a separate PR since it is a non-trivial thing.

tests/conftest.py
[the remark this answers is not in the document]
RyanMarten: will remove

vutrung96 (approved):
approved! thanks for the PR

PR #311, opened 2025-01-07
Add error handling and retries for batch requests

### LINKED ISSUE #279: openai.APIConnectionError when using batch mode not caught
```
-12-19_07-40-49_467973_1/runtime_resources/conda/01b738786d663797e9dc09585bc0cbdc81b2bf5f/lib/python3.10/site-packages/openai/_base_client.py", line 1666, in _retry_request
    return await self._request(
  File "/tmp/ray/session_2024-12-19_07-40-49_467973_1/runtime_resources/conda/01b738786d663797e9dc09585bc0cbdc81b2bf5f/lib/python3.10/site-packages/openai/_base_client.py", line 1606, in _request
    raise APIConnectionError(request=request) from err
openai.APIConnectionError: Connection error.
```

From the docs
https://help.openai.com/en/articles/6897191-apiconnectionerror

>  An APIConnectionError indicates that your request could not reach our servers or establish a secure connection. This could be due to a network issue, a proxy configuration, an SSL certificate, or a firewall rule.
>  
> If you encounter an APIConnectionError, please try the following steps:
> 1. Check your network settings and make sure you have a stable and fast internet connection. You may need to switch to a different network, use a wired connection, or reduce the number of devices or applications using your bandwidth.
> 2. Check your proxy configuration and make sure it is compatible with our services. You may need to update your proxy settings, use a different proxy, or bypass the proxy altogether.
> 3. Check your SSL certificates and make sure they are valid and up-to-date. You may need to install or renew your certificates, use a different certificate authority, or disable SSL verification.
> 4. Check your firewall rules and make sure they are not blocking or filtering our services. You may need to modify your firewall settings.
> 
> If the issue persists, contact our support team and provide them with the following information:
> - The model you were using
> - The error message and code you received
> - The request data and headers you sent
> - The timestamp and timezone of your request
> - Any other relevant details that may help us diagnose the issue

Is this something we can catch and then handle more robustly?

shreyaspimpalgaonkar:
Fixes #279

### DIFF
--- a/src/bespokelabs/curator/request_processor/batch/anthropic_batch_request_processor.py
+++ b/src/bespokelabs/curator/request_processor/batch/anthropic_batch_request_processor.py
@@ -2,7 +2,7 @@ import logging
 import litellm
 import instructor
 
-from anthropic import AsyncAnthropic
+from anthropic import AsyncAnthropic, APIConnectionError
 from anthropic.types.messages import MessageBatch
 from anthropic.types.messages import MessageBatchRequestCounts
 from anthropic.types.shared.not_found_error import NotFoundError
@@ -261,13 +261,16 @@ class AnthropicBatchRequestProcessor(BaseBatchRequestProcessor):
             response_cost=cost,
         )
 
-    async def submit_batch(self, requests: list[dict], metadata: dict) -> GenericBatch:
+    async def submit_batch(
+        self, requests: list[dict], metadata: dict, num_retries: int = 3
+    ) -> GenericBatch:
         """
         Handles the complete batch submission process.
 
         Args:
             requests (list[dict]): List of API-specific requests to submit
             metadata (dict): Metadata to be included with the batch
+            num_retries (int): The number of retries to attempt
 
         Returns:
             Batch: The created batch object from OpenAI
@@ -276,7 +279,21 @@ class AnthropicBatchRequestProcessor(BaseBatchRequestProcessor):
             - Updates tracker with submitted batch status
         """
         async with self.semaphore:
-            batch = await self.client.messages.batches.create(requests=requests)
+            for retry_id in range(num_retries + 1):
+                try:
+                    batch = await self.client.messages.batches.create(requests=requests)
+                    break
+
+                except APIConnectionError as e:
+                    logger.error(
+                        f"Failed to connect to Anthropic API. Please check your internet connection."
+                    )
+                    if retry_id == num_retries:
+                        raise e
+                    else:
+                        logger.error(
+                            f"Retrying batch submission (attempt {retry_id+1}/{num_retries})"
+                        )
             return self.parse_api_specific_batch_object(
                 batch, request_file=metadata["request_file"]
             )
--- a/src/bespokelabs/curator/request_processor/batch/openai_batch_request_processor.py
+++ b/src/bespokelabs/curator/request_processor/batch/openai_batch_request_processor.py
@@ -5,7 +5,7 @@ import logging
 import litellm
 
 from typing import Optional
-from openai import AsyncOpenAI, NotFoundError
+from openai import AsyncOpenAI, NotFoundError, APIConnectionError
 from openai.types.batch import Batch
 from openai.types.file_object import FileObject
 from openai.types.batch_request_counts import BatchRequestCounts
@@ -216,7 +216,7 @@ class OpenAIBatchRequestProcessor(BaseBatchRequestProcessor, OpenAIRequestMixin)
 
         return request
 
-    async def upload_batch_file(self, file_content: bytes) -> FileObject:
+    async def upload_batch_file(self, file_content: bytes, num_retries: int = 3) -> FileObject:
         """
         Uploads a batch file to OpenAI and waits until ready.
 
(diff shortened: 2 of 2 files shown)


src/bespokelabs/curator/request_processor/batch/anthropic_batch_request_processor.py
```diff
@@ -276,7 +279,21 @@ async def submit_batch(self, requests: list[dict], metadata: dict) -> GenericBat
             - Updates tracker with submitted batch status
         """
         async with self.semaphore:
-            batch = await self.client.messages.batches.create(requests=requests)
+            for retry_id in range(num_retries + 1):
```
adamoptimizer, line 282: This retry logic could goto higher level api.
Because the retry will be needed in all apis (Openai, Anthropic, etc)
so we need to create a custom expection here and catch it in BaseBatchProcessor or something and this way it could scale to all apis.

WDYT? @vutrung96 @shreyaspimpalgaonkar

adamoptimizer (changes requested):

src/bespokelabs/curator/request_processor/batch/anthropic_batch_request_processor.py
vutrung96: +1 could you please move this to the higher level [base_batch_request_processor.py](https://github.com/bespokelabsai/curator/blob/d6244c43260b748cc78440dec8957f5a63fc071e/src/bespokelabs/curator/request_processor/batch/base_batch_request_processor.py) @shreyaspimpalgaonkar?

RyanMarten: Also @shreyaspimpalgaonkar you can actually pass some of the retry logic into the client via http client. This may be cleaner? 

Check out (openai)
https://github.com/openai/openai-python?tab=readme-ov-file#handling-errors
https://github.com/openai/openai-python?tab=readme-ov-file#retries
https://github.com/openai/openai-python?tab=readme-ov-file#configuring-the-http-client

(anthropic - very similar / same)
https://github.com/anthropics/anthropic-sdk-python?tab=readme-ov-file#handling-errors
https://github.com/anthropics/anthropic-sdk-python?tab=readme-ov-file#retries
https://github.com/anthropics/anthropic-sdk-python?tab=readme-ov-file#configuring-the-http-client

vutrung96: also why do we only retry for APIConnectionError? we can also just retry for all errors?

that will simplify this a lot since we only need to configure max_retries for that behavior?

vutrung96: Stainless only retries for a specific set of retriable errors (API Connection behing one of them): https://app.stainlessapi.com/docs/guides/configure#retries.

shreyaspimpalgaonkar: So we already have a max_retries argument in curator.LLM(), and we initialize the OpenAI and Anthropic clients using those parameters. So internally, the client is already retrying `max_retries` times and we don't need to retry the request here. 

https://github.com/openai/openai-python/blob/728100cf7a7aff90bbe274e593bc726030e392ec/src/openai/_base_client.py#L1001

So, we just need to copy paste this logic 
https://github.com/openai/openai-python?tab=readme-ov-file#handling-errors

Note that even though it may look similar, openai and anthropic don't return the same exceptions or call the same function. So the logic needs to be written in respective classes, or we write an exception mapping between the base and derived classes.

vutrung96: if we are already retrying via Stainless, then why do we need this PR to manually retry? :D

vutrung96: or is it that OpenAI/Anthropic Stainless clients don't retry APIConnection errors (this is weird to me)?

shreyaspimpalgaonkar: Yes, we don't need to manually retry. I wasn't aware before that we were using this functionality (also, I added that in because the issue mentions that we should manually retry). 

in the above link, you can see that Openai and anthropic clients retry APIConnectionerrors. 

We just need to catch and log the errors cleanly.

RyanMarten: Yes, I believe that the errors get thrown after the max number of retries @vutrung96 @shreyaspimpalgaonkar

vutrung96: So it seems the solution is to raise max_retries? Although i think I have i set to 50, so i don't understand how it could fail 50 times in a row if its an ephemeral network error.

RyanMarten: The default is now 10 https://github.com/bespokelabsai/curator/blob/65892b49dcf2d944bdcdfab130d23f985fd15b0a/src/bespokelabs/curator/request_processor/config.py#L13

vutrung96: ok, I think we can keep it at 10 and just undo this PR and close the bug. iiuc we might have just failed 10 times in a row in the bug.

vutrung96: @RyanMarten re. which error is retried, you can see the logic here

https://github.com/openai/openai-python/blob/728100cf7a7aff90bbe274e593bc726030e392ec/src/openai/_base_client.py#L992

specifically, APIConnectionError is only raised after retries are exhausted: 

https://github.com/openai/openai-python/blob/728100cf7a7aff90bbe274e593bc726030e392ec/src/openai/_base_client.py#L1027

PR #329, opened 2025-01-08
Relax VLLM version requirements

vutrung96:
Currently the versioning is way too strict and pinned to a specific version. We should allow everything from 0.6.0 and above.

0.6.3 should support all the features we want:
* LLM constructor arguments (used [here]( https://github.com/bespokelabsai/curator/blob/2f3f849e83f6c707279253952b6bac5ccf8f6819/src/bespokelabs/curator/request_processor/offline/vllm_offline_request_processor.py#L40)): all these constructor arguments are support in 0.6.3 https://docs.vllm.ai/en/v0.6.3/dev/offline_inference/llm.html
* GuidedDecodingParams (used [here]( https://github.com/bespokelabsai/curator/blob/2f3f849e83f6c707279253952b6bac5ccf8f6819/src/bespokelabs/curator/request_processor/offline/vllm_offline_request_processor.py#L95)): first supported in [v0.6.3](https://github.com/vllm-project/vllm/blob/v0.6.3/vllm/sampling_params.py)


RyanMarten (approved):
thanks

GeorgiosSmyrnis:
Thanks @vutrung96 !

PR #330, opened 2025-01-09
Fix circular import

GeorgiosSmyrnis:
This change seems needed to fix a circular import issue when loading curator.


vutrung96 (approved):
Thanks George!

PR #335, opened 2025-01-10
feat: add ability to disable caching via CURATOR_DISABLE_CACHE

vutrung96:
when CURATOR_DISABLE_CACHE is turned on, we will randomize the hash for all datasets, which will lead to every LLM call creating in a new dataset.

the benefit: we don't destroy any user-created dataset, and users can view all these datasets in curator-viewer
the downside: the dataset might build up in cache causing out of disk problem, but we can warn user about this in our documentation


src/bespokelabs/curator/llm/llm.py
```diff
@@ -110,8 +110,32 @@ def __init__(
...
+            fingerprint = xxh64(os.urandom(8)).hexdigest()
+        else:
+            prompt_func_hash = _get_function_hash(self.prompt_formatter.prompt_func)
+
+            fingerprint_str = "_".join(
+                [
+                    str(dataset_hash),
+                    str(prompt_func_hash),
+                    str(self.prompt_formatter.model_name),
+                    str(self.prompt_formatter.response_format.model_json_schema() if self.prompt_formatter.response_format else "text"),
```
adamoptimizer, line 124: this is permute equivariant,

fingerprint changes on order of repose_format

tests/test_caching.py
```diff
@@ -203,3 +203,26 @@ def test_func():
         logger.debug(f"  hash2: {hash2}")
 
         assert hash1 == hash2, "Identical functions should produce the same hash"
+
+
+def test_disable_cache(tmp_path):
```
adamoptimizer, line 208: may be vcr this?
else,
should we patch the llm run and see if it bypasses the cache safely, twice?

adamoptimizer (changes requested):
LGTM, just 2 comments,
Thanks for adding this feature :)

src/bespokelabs/curator/llm/llm.py
vutrung96: spoke offline we agreed to fix this later

tests/test_caching.py
vutrung96: done! could you ptal?

adamoptimizer (approved):
Thanks, LGTM

PR #343, opened 2025-01-11
Rynam/batch retry fix

adamoptimizer:


tests/integrations/test_all.py
RyanMarten, line 1: Instead of a separate _FAILED_BATCH_BACKENDS use _BATCH_BACKENDS. Also please include `anthropic` in the list.

tests/integrations/openai/fixtures/failed_request_batch_resume.yaml
```diff
@@ -0,0 +1,680 @@
...
+      - '0'
+      x-stainless-runtime:
+      - CPython
+      x-stainless-runtime-version:
+      - 3.11.7
+    method: GET
+    uri: https://api.openai.com/v1/files/file-PWB1SHg1dfa4aHwozaauxd/content
+  response:
+    body:
+      string: '{"id": "batch_req_678229d0515081908626650ba40680ce", "custom_id": "2",
```
RyanMarten, line 626: Where is the failure here?

tests/integrations/openai/fixtures/.test_cache.zip
RyanMarten, line 1: What is in here?

src/bespokelabs/curator/request_processor/batch/base_batch_request_processor.py
```diff
@@ -358,6 +358,12 @@ async def submit_batch_from_request_file(
         metadata = {"request_file": request_file}
         requests = self.requests_from_generic_request_file(request_file, completed_request_ids)
         batch = await self.submit_batch(requests, metadata)
+
+        # Change state of batch if submitted and partial batch.
+        if len(completed_request_ids) > 0:
```
RyanMarten, line 363: My view is to not add this for a couple reasons:
1. potentially all requests errored out low funds or expiration so completion_request_ids could be 0 and we could be resubmitting
2. we now have to add new logic everywhere we are referring to `self.tracker.n_submitted_batches` or `self.tracker.submitted_batches` or `self.tracker.n_submitted_finished_or_downloaded_batches` (or update these in the tracker and maybe track separately?). Submitted is a super set of resubmitted. 

I made the decision to keep the number of statuses low in the tracker because it is already a bit difficult to debug and understand and I think it is ok to not mark as resubmitted.

RyanMarten (changes requested):
Thanks for adding the test, I suggested removing RESUBMITTED status for now

PR #350, opened 2025-01-12
test: add anthropic integration tests

adamoptimizer:
extended the current tests with anthropic backend.


tests/integrations/test_all.py
```diff
@@ -160,8 +164,8 @@ def _run_loop(func):
 
             mocked_run_loop.side_effect = _run_loop
             with pytest.raises(ValueError):
-                helper.create_basic(temp_working_dir, mock_dataset, batch=True)
-        mocked_run_loop.side_effect = run_in_event_loop
+                _reload_batch_patch_deps()
```
vutrung96, line 167: we need to reload because we patch poll_and_process_batch to fail? why didn't we need it before?

tests/integrations/test_all.py
```diff
@@ -183,6 +189,7 @@ def test_failed_request_in_batch_resume(caplog, temp_working_dir, mock_dataset):
     temp_working_dir, backend, vcr_config = temp_working_dir
```
vutrung96, line 189: I'm curious how you managed to simulate failed requests in a batch. Was it a lot of work?

vutrung96 (approved):
Thanks! just some questions for my knowledge

tests/integrations/test_all.py
adamoptimizer: - removed some responses,
- updated batch_objects with 2 succeed 1 failed 3 total
- remove cached arrow file.

PR #358, opened 2025-01-12
fix: broken vllm offline processor + add test

vutrung96:
VLLM offline processor was broken due to some recent changes.

I added a test to prevent regressions like this in the future, basically i'm running the test_basic test and then recording all the output similar to vcr, but i had to manually mock things out.


src/bespokelabs/curator/request_processor/offline/base_offline_request_processor.py
RyanMarten, line 1: thanks for updating this!

RyanMarten (approved):
LGTM in terms of code changes. 

The CI is saying one of the integration tests is failing with 
```
FAILED tests/integrations/test_all.py::test_basic_offline[temp_working_dir0] - ValueError: mutable default <class 'bespokelabs.curator.status_tracker.offline_status_tracker.System'> for field system is not allowed: use default_factory
```

Fix and merge

src/bespokelabs/curator/status_tracker/offline_status_tracker.py
```diff
@@ -44,7 +44,7 @@ class OfflineStatusTracker:
     time_finished: datetime.datetime = None
     finished_successfully: bool = False
     num_total_requests: int = 0
-    system: System = System()
+    system: System = field(default_factory=System)
```
RyanMarten, line 47: Issue seems to be here

vutrung96: fixed!

src/bespokelabs/curator/request_processor/offline/base_offline_request_processor.py
vutrung96: np!

PR #359, opened 2025-01-13
Add fix_json check

GeorgiosSmyrnis:
This fixes an issue where `fix_json` is applied even if structured output is not requested in the VLLM offline request processor.

### DIFF
--- a/src/bespokelabs/curator/request_processor/config.py
+++ b/src/bespokelabs/curator/request_processor/config.py
@@ -139,7 +139,6 @@ class OfflineBackendParams(BaseBackendParams, total=False):
     enforce_eager: t.Optional[bool]
     max_model_length: t.Optional[int]
     max_tokens: t.Optional[int]
-    min_tokens: t.Optional[int]
     gpu_memory_utilization: t.Optional[float]
     batch_size: t.Optional[int]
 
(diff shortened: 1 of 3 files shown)


GeorgiosSmyrnis:
Result seems to be as expected, so edited the hash for the result expected in the test as well

src/bespokelabs/curator/request_processor/config.py
```diff
@@ -139,6 +139,7 @@ class OfflineBackendParams(BaseBackendParams, total=False):
     enforce_eager: t.Optional[bool]
     max_model_length: t.Optional[int]
     max_tokens: t.Optional[int]
+    min_tokens: t.Optional[int]
```
vutrung96, line 142: why do we need to add min_tokens here? where are we using it?

vutrung96 (changes requested):
mostly looks good apart from the weird min_tokens change? did you accidentally add it?

src/bespokelabs/curator/request_processor/config.py
GeorgiosSmyrnis: I thought that it needed to be added because of validator (since it is an argument of `OfflineRequestProcessorConfig`). I'll remove it

vutrung96 (approved):

PR #362, opened 2025-01-14
Fix_json adds curly braces

marianna13:


RyanMarten:
Fixed in #359 

Slightly different from what you have here. 
We can still merge in the other bug fix.

PR #371, opened 2025-01-16
Allow lists of messages as simple input

RyanMarten:
Allowing for 

Right now we allow 
```
llm("hi")
```
```
llm(["hi", "tell me a joke"])
```

I'm looking to also allow 
```
llm(list[messages])
```

like
```
llm([[{'role': 'user', 'content': 'What are the names of some famous actors that started their careers on Broadway?'}], [{'role': 'user', 'content': 'How did US states get their names?'}]])
```

### DIFF
--- a/src/bespokelabs/curator/llm/llm.py
+++ b/src/bespokelabs/curator/llm/llm.py
@@ -7,6 +7,7 @@ from datetime import datetime
 from io import BytesIO
 from typing import TYPE_CHECKING, Any, Dict, Iterable, Optional, Type, TypeVar
 
+from datasets import Dataset
 from pydantic import BaseModel
 from xxhash import xxh64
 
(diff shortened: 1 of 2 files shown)


RyanMarten:
```
from bespokelabs import curator
llm = curator.LLM("gpt-4o-mini")
llm([[{'role': 'user', 'content': 'What are the names of some famous actors that started their careers on Broadway?'}], [{'role': 'user', 'content': 'How did US states get their names?'}]])
```

RyanMarten:
How about a single message? 
```
from bespokelabs import curator
llm = curator.LLM("gpt-4o-mini")
llm([{'role': 'user', 'content': 'What are the names of some famous actors that started their careers on Broadway?'}])
```
This breaks it.

RyanMarten:
Maybe it gets too confusing then since an iterator of dataset rows looks similar to a message --> both are like lists of dict

src/bespokelabs/curator/llm/llm.py
```diff
@@ -175,11 +175,14 @@ def __call__(
...
+        elif isinstance(dataset, list) and all(isinstance(item, dict) and "role" in item and "content" in item for item in dataset):
+            # A list of messages is converted to a dataset with a single row
+            dataset = Dataset.from_list([{"prompt": dataset}])
         elif not isinstance(dataset, Dataset) and dataset is not None:
-            # Wrap the iterable in a generator to convert strings to dictionaries
+            # Wrap the iterable in a generator, the prompt is expected to be a prompt string or a list of messages
             def wrapped_iterable():
                 for input in dataset:
-                    if isinstance(input, str):
+                    if isinstance(input, str) or isinstance(input, list):
```
vutrung96, line 185: this check on line 185 and the check on line 178 are the same check (whether the prompt is a list of messages) right? but the logic is written differently?

we should add function to unify this logic (maybe `is_messages_list`) and test it somehow. best way to test would probably be to do an end-to-end test with a cassette or mock out the client, but short of that, we should at least test that this logic works properly in a unit test.

RyanMarten: adding now

RyanMarten: addressed :)

vutrung96 (approved):
thanks!

Issue #377, opened 2025-01-17
Add support for relying on max_parallel_requests instead of max_requests_per_minute and max_tokens_per_minute


alpayariyak:

vutrung96:
@alpayariyak thanks for suggesting this feature! could you elaborate more on your use case, e.g. which provider requires this form of rate limiting?

RyanMarten:
related: #253

RyanMarten:
Example provider is `deepinfra` - https://deepinfra.com/docs/advanced/rate-limits. `max_concurrent_requests`

Have you seen this for other providers @alpayariyak?

alpayariyak:
Thank you all for the quick response and PR! Personally, I need this for generating data with vLLM, as it's hard to granularly control concurrent requests with current limiters, so I either end up sending too few requests or too many (which leads to countless aborted requests).

If it's of any help, here's how I added it in my fork, haven't tested it with using the previous limiter methods: https://github.com/alpayariyak/curator/commit/d5dd941563c382120bace90fc8c1ea534267f8a3

RyanMarten:
@alpayariyak We just merged in #378 which should address this for you!
Please let us know if you run into any issues!

Issue #253, opened 2024-12-13
Rate limiting abstraction that captures multiple types


RyanMarten:
Different types of rate limits

requests per minute / tokens per minute

active connections (deep infra)

RyanMarten:
Simple way to deal with active connections is just use a sempaphore

RyanMarten:
`max_concurrent_requests`

PR #378, opened 2025-01-18
feat: support max parallel request processor

adamoptimizer:
This pr supports max parallel requests rate limit

This rate limit can be found in Deep Infra models.

https://github.com/bespokelabsai/curator/issues/377

### DIFF
--- a/src/bespokelabs/curator/request_processor/config.py
+++ b/src/bespokelabs/curator/request_processor/config.py
@@ -79,11 +79,13 @@ class OnlineRequestProcessorConfig(RequestProcessorConfig):
         max_tokens_per_minute: Maximum number of tokens allowed per minute
         max_input_tokens_per_minute: Maximum number of input tokens allowed per minute
         max_output_tokens_per_minute: Maximum number of output tokens allowed per minute
+        max_concurrent_requests: Maximum number of concurrent requests
         seconds_to_pause_on_rate_limit: Duration to pause when rate limit is hit
         invalid_finish_reasons: List of api finish reasons which are considered failed.
     """
 
     max_requests_per_minute: int | None = Field(default=None, gt=0)
+    max_concurrent_requests: int | None = Field(default=None, gt=0)
     max_tokens_per_minute: int | None = Field(default=None, gt=0)
     max_input_tokens_per_minute: int | None = Field(default=None, gt=0)
     max_output_tokens_per_minute: int | None = Field(default=None, gt=0)
--- a/src/bespokelabs/curator/request_processor/online/base_online_request_processor.py
+++ b/src/bespokelabs/curator/request_processor/online/base_online_request_processor.py
@@ -1,7 +1,7 @@
 """Base class for online request processors that make real-time API calls.
 
 This module provides the core functionality for making API requests in real-time,
-handling rate limiting, retries, and parallel processing.
+handling rate limiting, retries, and concurrent processing.
 """
 
 import asyncio
@@ -9,6 +9,7 @@ import datetime
 import json
 import logging
 import time
+import typing as t
 from abc import ABC, abstractmethod
 from collections import deque
 from dataclasses import dataclass, field
@@ -75,14 +76,37 @@ class BaseOnlineRequestProcessor(BaseRequestProcessor, ABC):
         self.header_based_max_requests_per_minute = None
         self.header_based_max_tokens_per_minute = None
 
+        self.manual_max_concurrent_requests = config.max_concurrent_requests
+        self.header_based_max_concurrent_requests = None
+
         # The rich.Console used for the status tracker, only set for testing
         self._tracker_console = None
         self._output_tokens_window = deque(maxlen=_MAX_OUTPUT_MVA_WINDOW)
+        self._semaphore = None
+        if self.max_concurrent_requests is not None:
+            self._semaphore = asyncio.Semaphore(t.cast(int, self.max_concurrent_requests))
 
     @property
     def backend(self) -> str:
         """Backend property."""
         return "base"
+
+    @property
+    def max_concurrent_requests(self) -> int | None:
+        """Gets the maximum concurrent requests rate limit.
+
+        Returns the manually set limit if available, falls back to header-based limit,
+        or uses default value as last resort.
+        """
+        if self.manual_max_concurrent_requests:
+            logger.info(f"Manually set max_concurrent_requests to {self.manual_max_concurrent_requests}")
+            return self.manual_max_concurrent_requests
+
+        elif self.header_based_max_concurrent_requests:
+            logger.info(f"Automatically set max_concurrent_requests to {self.header_based_max_concurrent_requests}")
+            return self.header_based_max_concurrent_requests
+        else:
+            return None
 
     @property
     def max_requests_per_minute(self) -> int:
@@ -207,6 +231,9 @@ class BaseOnlineRequestProcessor(BaseRequestProcessor, ABC):
             logger.warn(f"Pausing for {int(remaining_seconds_to_pause)} seconds")
             await asyncio.sleep(remaining_seconds_to_pause)
 
+    def free_capacity(self, tracker, tokens):
+        """Free blocked capacity."""
+
     async def process_requests_from_file(
         self,
         generic_request_filepath: str,
@@ -226,8 +253,8 @@ class BaseOnlineRequestProcessor(BaseRequestProcessor, ABC):
             max_tokens_per_minute=self.max_tokens_per_minute,
         )
 
-        # Resume if a response file exists
         completed_request_ids = self.validate_existing_response_file(response_file)
+        # Resume if a response file exists
 
         # Count total requests
         status_tracker.num_tasks_already_completed = len(completed_request_ids)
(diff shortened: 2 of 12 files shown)


RyanMarten:
related #253

src/bespokelabs/curator/request_processor/online/base_online_request_processor.py
```diff
@@ -61,18 +61,29 @@ class BaseOnlineRequestProcessor(BaseRequestProcessor, ABC):
...
         self.header_based_max_requests_per_minute = None
         self.header_based_max_tokens_per_minute = None
 
-        # The rich.Console used for the status tracker, only set for testing
-        self._tracker_console = None
+        self.manual_max_parallel_requests = config.max_parallel_requests
+        self.default_max_parallel_requests = 200
+        self.header_based_max_parallel_requests = None
+
+    def _dynamic_mixin_loader(self, strategy):
```
RyanMarten, line 82: This is a little complicated to me

src/bespokelabs/curator/request_processor/online/base_online_request_processor.py
```diff
@@ -322,103 +442,144 @@ async def process_requests_from_file(
...
+    @property
+    def max_parallel_requests(self: BaseOnlineRequestProcessor) -> int:
+        """Gets the maximum parallel requests rate limit.
+
+        Returns the manually set limit if available, falls back to header-based limit,
+        or uses default value as last resort.
+        """
+        if self.manual_max_parallel_requests:
+            logger.info(f"Manually set max_parallel_requests to {self.manual_max_parallel_requests}")
+            return self.manual_max_parallel_requests
```
RyanMarten, line 456: @vutrung96 made this comment before, should be `concurrent` instead of `parallel`

RyanMarten:
Briefly took a look. 
The implementation looks a bit more complicated that anticipated. 

We already have a `concurrent` limit in the batch request processor and it simply uses `asyncio.Semaphore`

https://github.com/bespokelabsai/curator/blob/83d59645a0306e19b279e5376a3cf84326b7b15e/src/bespokelabs/curator/request_processor/batch/base_batch_request_processor.py#L90

https://github.com/bespokelabsai/curator/blob/83d59645a0306e19b279e5376a3cf84326b7b15e/src/bespokelabs/curator/request_processor/batch/anthropic_batch_request_processor.py#L256-L271

This was my first though in #253 on how to solve it. 

Could we do the same for online here by adding `async with self.semaphore`

https://github.com/bespokelabsai/curator/blob/83d59645a0306e19b279e5376a3cf84326b7b15e/src/bespokelabs/curator/request_processor/online/base_online_request_processor.py#L262-L270

Does that approach have an issue I'm not seeing?

adamoptimizer:
@RyanMarten 
This is still in draft!

adamoptimizer:
@RyanMarten check this out, Im referencing https://deepinfra.com/docs/advanced/rate-limits
for rate limits shared by Trung.

adamoptimizer:
> @adamoptimizer Briefly took a look. The implementation looks a bit more complicated that anticipated.
> 
> We already have a `concurrent` limit in the batch request processor and it simply uses `asyncio.Semaphore`
> 
> https://github.com/bespokelabsai/curator/blob/83d59645a0306e19b279e5376a3cf84326b7b15e/src/bespokelabs/curator/request_processor/batch/base_batch_request_processor.py#L90
> 
> Example use in `submit_batch`
> 
> https://github.com/bespokelabsai/curator/blob/83d59645a0306e19b279e5376a3cf84326b7b15e/src/bespokelabs/curator/request_processor/batch/anthropic_batch_request_processor.py#L269-L271
> 
> This was my first though in #253 on how to solve it.
> 
> Could we do the same for online here by adding `async with self.semaphore`. We can allow the user to set all three (`concurrent`, `rpm`, `tpm`) if that is really something they want, instead of having it be two distinct classes / strategies? Setting one to `inf` essentially disables that rate limit.
> 
> https://github.com/bespokelabsai/curator/blob/83d59645a0306e19b279e5376a3cf84326b7b15e/src/bespokelabs/curator/request_processor/online/base_online_request_processor.py#L262-L270
> 
> Does that approach have an issue I'm not seeing?

No, You are correct, will use semaphore to limit concurrency, 
The above code was more debuggable!

src/bespokelabs/curator/request_processor/online/base_online_request_processor.py
adamoptimizer: Ignore this, 
Since we decided to use rpm, tpm and concurrent at same time, no need for this.

src/bespokelabs/curator/request_processor/config.py
```diff
@@ -76,10 +76,12 @@ class OnlineRequestProcessorConfig(RequestProcessorConfig):
     Attributes:
         max_requests_per_minute: Maximum number of requests allowed per minute
         max_tokens_per_minute: Maximum number of tokens allowed per minute
+        max_concurrent_requests: Maximum number of concurrent requests
         seconds_to_pause_on_rate_limit: Duration to pause when rate limit is hit
     """
 
     max_requests_per_minute: int | None = Field(default=None, gt=0)
+    max_concurrent_requests: int | None = Field(default=None, gt=0)
```
RyanMarten, line 84: should the default be `inf` here?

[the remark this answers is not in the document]
RyanMarten: Or actually it can be None and we can just replace
`if self.max_concurrent_requests != float("inf"):`
with 
`if self.max_concurrent_requests is not None:
(for all of the `max_rpm` `max_concurrent` `max_tpm`) and also where we set TCP connection limit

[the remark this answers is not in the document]
RyanMarten: if we use none, would this break anything else anywhere in our code?

src/bespokelabs/curator/request_processor/online/base_online_request_processor.py
```diff
@@ -213,10 +316,12 @@ async def process_requests_from_file(
         # Initialize trackers
         queue_of_requests_to_retry: asyncio.Queue[APIRequest] = asyncio.Queue()
         status_tracker = OnlineStatusTracker()
+        token_estimate = float("inf")
```
RyanMarten, line 319: is this right?

src/bespokelabs/curator/request_processor/config.py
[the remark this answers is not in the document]
RyanMarten: As I am reading onwards, the cascading logic of defaults and how they are set in the config, then the base processor, and finally in litellm processor is a little bit confusing to keep track of. I may have added to this confusion by adding this idea of `inf`. Maybe that wasn't necessary.

src/bespokelabs/curator/request_processor/online/litellm_online_request_processor.py
```diff
@@ -39,12 +41,55 @@ class LiteLLMOnlineRequestProcessor(BaseOnlineRequestProcessor):
...
+        or uses default value as last resort.
+        """
+        if self._concurrency_only_rate_limited:
+            return float("inf")
+        else:
+            return super().max_tokens_per_minute
+
+    @property
+    def _concurrency_only_rate_limited(self):
+        provider = self.config.model.split("/")[0]
```
RyanMarten, line 90: I feel like what we are ending up with here is a little like https://github.com/BerriAI/litellm/blob/main/model_prices_and_context_window.json

Maybe we should just store a JSON with default values for all three of the limits. 
That way we avoid extra logic and just read in directly
And also that way, people in the community can easily add extra. 

For example, we should store `200` for deepinfra concurrent (since that is the starting amount I believe). 
And similarly, for gemini, store the values for the basic paid account (since they don't do automatic rate limits).

And this _CONCURRENT_ONLY_RATELIMIT_PROVIDERS would basically be captured by this since those providers would have TPM and RPM set to none in the json. 
technically, there is nothing stopping the user from adding TPM and RPM rate limiting since this is happening client side.

RyanMarten:
Not sure how to write tests for this one

src/bespokelabs/curator/request_processor/online/litellm_online_request_processor.py
RyanMarten: Added an issue about this here: #388

adamoptimizer: agreed

src/bespokelabs/curator/request_processor/online/base_online_request_processor.py
```diff
@@ -64,21 +65,45 @@ class BaseOnlineRequestProcessor(BaseRequestProcessor, ABC):
...
+        """Gets the maximum concurrent requests rate limit.
+
+        Returns the manually set limit if available, falls back to header-based limit,
+        or uses default value as last resort.
+        """
+        if self.manual_max_concurrent_requests:
+            logger.info(f"Manually set max_concurrent_requests to {self.manual_max_concurrent_requests}")
+            return self.manual_max_concurrent_requests
+
+        elif self.header_based_max_concurrent_requests:
```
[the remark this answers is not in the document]
adamoptimizer: yes, I tried to do it and couldnt find the headers, lets keep it

src/bespokelabs/curator/request_processor/config.py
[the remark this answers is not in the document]
adamoptimizer: `inf` just represents unlimited capacity, 
None would work as well.

do you want me to change it to `None`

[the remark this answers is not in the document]
adamoptimizer: @RyanMarten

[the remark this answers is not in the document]
adamoptimizer: I have made it None

src/bespokelabs/curator/request_processor/online/base_online_request_processor.py
adamoptimizer: removed

RyanMarten (approved):
LGTM! Thanks @adamoptimizer

Issue #375, opened 2025-01-17
Allow user to specify which `finish_reason`'s to retry / fail on


RyanMarten:
Right now this is hard coded in the litellm processor: https://github.com/bespokelabsai/curator/blob/83d59645a0306e19b279e5376a3cf84326b7b15e/src/bespokelabs/curator/request_processor/online/litellm_online_request_processor.py#L262-L266

This should be abstracted at the highest level so it is true for all request processors. 

For example, in evalchemy, we set max_tokens and actually want truncated responses to be considered successful
- https://github.com/mlfoundations/evalchemy/pull/51

I'm thinking that we can pass in `invalid_finish_reasons` to `backend_params`.

adamoptimizer:
@RyanMarten you mean bypassing invalid finish error with a flag in `backend_params`

RyanMarten:
I mean actually specifying the list of invalid finish reasons @adamoptimizer

RyanMarten:
- https://github.com/BerriAI/litellm/pull/7264

PR #380, opened 2025-01-19
ref: make `invalid_finish_reasons` configurable

adamoptimizer:
https://github.com/bespokelabsai/curator/issues/375

### DIFF
--- a/src/bespokelabs/curator/request_processor/online/base_online_request_processor.py
+++ b/src/bespokelabs/curator/request_processor/online/base_online_request_processor.py
@@ -9,6 +9,7 @@ import datetime
 import json
 import logging
 import time
+import typing as t
 from abc import ABC, abstractmethod
 from collections import deque
 from dataclasses import dataclass, field
@@ -83,6 +84,11 @@ class BaseOnlineRequestProcessor(BaseRequestProcessor, ABC):
     def backend(self) -> str:
         """Backend property."""
         return "base"
+
+    @property
+    def invalid_finish_reasons(self) -> t.Sequence:
+        """List of api finish reasons which are considered failed."""
+        return self.config.invalid_finish_reasons
 
     @property
     def max_requests_per_minute(self) -> int:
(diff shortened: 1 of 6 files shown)


src/bespokelabs/curator/request_processor/online/base_online_request_processor.py
```diff
@@ -79,6 +80,11 @@ def backend(self) -> str:
         """Backend property."""
         return "base"
 
+    @property
+    def invalid_finish_reasons(self) -> t.Sequence:
```
RyanMarten, line 84: why do we need this instead of directly accessing via `self.config`?

src/bespokelabs/curator/request_processor/online/litellm_online_request_processor.py
```diff
@@ -50,6 +51,11 @@ def backend(self):
         """Backend property."""
         return "litellm"
 
+    @property
+    def invalid_finish_reasons(self) -> t.Sequence:
+        """List of api finish reasons which are considered failed."""
+        return ("length", "content_filter") or self.config.invalid_finish_reasons
```
RyanMarten, line 57: this default could go in 
`invalid_finish_reasons: list = Field(default_factory=list)`
in the config.py
I like the idea of all the defaults being in one place

RyanMarten (changes requested):
looks good, left some comments on the logic 

can take a look at the tests when they are added as well

src/bespokelabs/curator/request_processor/online/base_online_request_processor.py
adamoptimizer: invalid_finish_reasons are overloaded in Litellm class for ['length', 'content_filter'] defaults

src/bespokelabs/curator/request_processor/online/litellm_online_request_processor.py
adamoptimizer: Checked, They are valid for openai

src/bespokelabs/curator/request_processor/online/base_online_request_processor.py
RyanMarten: Those should just be the default for all online request processors - let's do it in config

kartik4949: sure lets do it

RyanMarten (approved):
LGTM!

PR #383, opened 2025-01-20
Organize dependencies

RyanMarten:
in the next release this means 
`pip install bespokelabs-curator[vllm]`
to use vllm

since the vllm llibrary is only used for offline vllm processing and is heavy (contains torch, etc.) I'm making it an extra

I also moved all the dev dependencies into the right section, and updated `make install` to download those
- vcr
- pytest
- isort


Makefile
```diff
@@ -25,7 +25,7 @@ clean:
 
 install:
 	@echo "Installing dependencies..."
-	poetry install
+	poetry install --with dev
```
adamoptimizer, line 28: much needed

pyproject.toml
```diff
@@ -23,26 +23,28 @@ python = "^3.10"
 pydantic = ">=2.9.2"
 datasets = "^3.0.2"
 instructor = "^1.6.3"
-pytest = "^8.3.3"
```
adamoptimizer, line 26: thanks for doin this

adamoptimizer (approved):
LGTM

PR #384, opened 2025-01-20
feat: add support for returning the full completions object + some changes to support deepseek models

vutrung96:
This PR
- Add support for returning the full completions object with 
- Make several changes to make deepseek models compatible: in particular, it turns out litellm does not fill out the reasoning_content field when it returns the response, so we need to route deepseek models to openai backend to get the raw response

I'll add more tests for this later.


madiator (approved):

RyanMarten (approved):
LGTM!

PR #406, opened 2025-01-25
ref: do not use request file info from batch metadata

adamoptimizer:
extend submission state with  `pre_schedule` 
skips exception raise in `upload_batch_file`
suppress warning in model.raw_dump


src/bespokelabs/curator/request_processor/batch/openai_batch_request_processor.py
```diff
@@ -234,12 +253,10 @@ async def upload_batch_file(self, file_content: bytes) -> FileObject:
...
         # Which results in a file not found error, so we briefly pause before checking the status
         await asyncio.sleep(1)
-
         try:
             batch_file_upload = await self.client.files.wait_for_processing(batch_file_upload.id)
-        except Exception as e:
-            logger.error(f"Error waiting for batch file to be processed: {e}")
-            raise e
+        except NotFoundError:
+            logger.warn("Could not check file upload status, file may not have been created!")
```
vutrung96, line 259: maybe can we skip this only for kluster.ai (can check base_url)? for openai batch this is actually an error right?

vutrung96 (approved):
just one minor suggestion

PR #407, opened 2025-01-25
Verifiers for Code

RyanMarten:


shreyaspimpalgaonkar:
Screenshot 2025-01-31 at 5 07 18 PM

tests/experimental/code_executor/test_code_execution.py
```diff
@@ -0,0 +1,54 @@
+# test with multiprocessing backend
```
vutrung96, line 1: move this out of experimental too?

vutrung96 (approved):
also remember to add this to function_hash to avoid recursion from pickling:

```
import sys


def _get_function_hash(func) -> str:
    """Get a hash of a function's source code."""
    sys.setrecursionlimit(10000)
```

PR #409, opened 2025-01-27
Remove upper bound from tiktoken

kartik4949:
This PR removes upper bound from tiktoken in pyproject toml


vutrung96 (approved):
thank you! why is integrating test failing?

Issue #410, opened 2025-01-27
Adding a curator.Dataset() object


shreyaspimpalgaonkar:
Currently, curator.LLM() returns a HF dataset object, we wish to change it to a curator.Dataset object.

A light wrapper for now, that overrides .push_to_hub() that uploads the dataset and adds these two things: 

1. a nice readme with a `Made with Curator` tag
2. code to easily replicate the data generation pipeline. this could be done by uploading the prompt and parse code snippets and overriding the load_dataset() function

RyanMarten:
I also think with a wrapper we can do nice things around caching non-llm steps that use datasets and return dataset but is not a dataset operation, so it is not picked up by the HF cache system (e.g. deduplication, decontamination that converts to pandas and back)

vutrung96:
yes agreed with @RyanMarten ! was just thinking of this yesterday.

there's a lot more curator-related context that we can pass around if we have our own dataset object.

PR #420, opened 2025-01-29
Override push_to_hub()

shreyaspimpalgaonkar:
Partly Solves #410 without creating the curator.Dataset object.

Creating a curator.Dataset object introduces additional logic + computational overhead. Member functions of HF datasets return dataset.arrow.Dataset object, which need to be cast to curator.Dataset for different cases, which is tricky. 

For now, our use case is a simple made_with_curator tag. This PR solves it.

Sample output:
https://huggingface.co/datasets/pimpalgaonkar/poems_test_3


RyanMarten (approved):
LGTM! We will fix the design of the card in post ;)

PR #422, opened 2025-01-29
Kluster.ai backend

RyanMarten:


src/bespokelabs/curator/request_processor/batch/openai_batch_request_processor.py
```diff
@@ -289,7 +287,7 @@ async def create_batch(self, batch_file_id: str, metadata: dict) -> Batch:
             batch = await self.client.batches.create(
                 input_file_id=batch_file_id,
                 endpoint="/v1/chat/completions",
-                completion_window="24h",
+                completion_window=self.config.completion_window,
```
shreyaspimpalgaonkar, line 290: Does the OpenAI batch API support a completion window other than 24h?

https://platform.openai.com/docs/guides/batch#3-creating-the-batch

RyanMarten: No, but Kluster does

RyanMarten: We can add config validation logic

shreyaspimpalgaonkar: A warning message that redirects users to use kluster AI could be nice too

CharlieJCJ:
fixed cost display issue in online example
image
image

src/bespokelabs/curator/request_processor/batch/openai_batch_request_processor.py
vutrung96: let's do this in a separate PR.

vutrung96 (approved):

PR #424, opened 2025-01-30
feat: add cost processor for litellm and external provider

kartik4949:


src/bespokelabs/curator/cost.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+class _LitellmCostProcessor:
+    def __init__(self, batch=False) -> None:
+        self.batch = batch
+
+    def cost(self, *args, completion_window="*", **kwargs):
+        import litellm
+
+        try:
```
vutrung96, line 14: can you check that the model is in model_cost instead of try-catching. example: https://github.com/bespokelabsai/curator/blob/73986c85390cc2a58316e9fbcbc5d068db728640/src/bespokelabs/curator/request_processor/online/openai_online_request_processor.py#L101C33-L101C51.

this gets rid of the annoying Provider List message in the terminal caused by looking up a non-existing model in LiteLLM.

image (6)

src/bespokelabs/curator/cost.py
```diff
@@ -0,0 +1,89 @@
...
+            model = args[0]
+            logging.warn(f"Could not retrieve cost for the model: {model}")
+        if self.batch:
+            cost_to_complete *= 0.5
+        return cost_to_complete
+
+
+# source: https://www.kluster.ai/#pricing
+_EXTERNAL_MODEL_COST_MAP = {
+    "klusterai/Meta-Llama-3.1-8B-Instruct-Turbo": {
```
vutrung96, line 27: we also need the cost for R1 here?

kartik4949: Im adding more models

vutrung96 (approved):

PR #429, opened 2025-01-31
fix: check model in litellm.cost instead of try catching

vutrung96:
The issue is that if the exception isn't litellm.exceptions.BadRequestError, we will error out. In fact this happened with DeepSeek-R1 on deepinfra
```
Exception: This model isn't mapped yet. model=deepseek-ai/DeepSeek-R1, custom_llm_provider=deepinfra. Add it here - 
https://github.com/BerriAI/litellm/blob/main/model_prices_and_context_window.json
```

We should just check explicitly if the model exists in the litellm's cost map instead of try-catching like this.


RyanMarten (approved):
LGTM

Issue #370, opened 2025-01-16
Request: Vary `generation_params` per request


RyanMarten:
Related to #328

richardzhuang0412:
Trying to bump this.

Use case: I'm replicating [BFCL v1 simple](https://huggingface.co/datasets/gorilla-llm/Berkeley-Function-Calling-Leaderboard#berkeley-function-calling-leaderboard) and I have generated a list of 100 function docs and 1 user question per function. Next step I would want to get the function call ground truth but it's unrealistic to support all 100 functions as `tools` parameter as this would confuse the model and potentially lead to wrong function calls. Thus I'd need `tools` to be varied per request.

Potential Solution: Maybe we can just add a preprocessing to the Dataset object passed in to `curator.LLM` that append a column of `generation_params` and when constructing the request get `generation_params` from the row?

PR #442, opened 2025-02-05
Added support for varying generation_params per request

richardzhuang0412:
Following #370, we want to enable the user to have a separate `generation_params` for each dataset entry.


kartik4949:
duplicate

PR #449, opened 2025-02-05
bump pyproject.toml and fix litellm version

shreyaspimpalgaonkar:


pyproject.toml
```diff
@@ -29,7 +29,7 @@ tqdm = "^4.67.0"
 matplotlib = "^3.9.2"
 nest-asyncio = "^1.6.0"
 rich = "^13.7.0"
-litellm = "1.55.4"
+litellm = ">=1.60.4"
```
vutrung96, line 32: maybe just use >= 1.55.4? why do we need to use this higher version?

vutrung96 (approved):

PR #463, opened 2025-02-07
fix: handle empty response

vutrung96:
DeepSeek can return empty responses during times of high load:
https://api-docs.deepseek.com/quick_start/rate_limit.

"However, please note that when our servers are under high traffic pressure, your requests may take some time to receive a response from the server. During this period, your HTTP request will remain connected, and you may continuously receive contents in the following formats:

Non-streaming requests: Continuously return empty lines
Streaming requests: Continuously return SSE keep-alive comments (: keep-alive)
"

From empirically testing sometimes they just close the connection and return 200 with an empty response. We need to handle this gracefully by detecting when response is None and raise an Exception. This will trigger a retry instead of simply error-ing out.

### DIFF
--- a/src/bespokelabs/curator/request_processor/online/openai_online_request_processor.py
+++ b/src/bespokelabs/curator/request_processor/online/openai_online_request_processor.py
@@ -221,7 +221,9 @@ class OpenAIOnlineRequestProcessor(BaseOnlineRequestProcessor, OpenAIRequestMixi
         ) as response_obj:
             response = await response_obj.json()
 
-            if "error" in response:
+            if response is None:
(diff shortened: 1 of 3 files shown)


madiator (approved):

src/bespokelabs/curator/request_processor/online/openai_online_request_processor.py
```diff
@@ -221,7 +221,9 @@ async def call_single_request(
         ) as response_obj:
             response = await response_obj.json()
```
shreyaspimpalgaonkar, line 222: if response_obj is None, that will throw an undesired error?

PR #466, opened 2025-02-07
test: raw prompt as list of dict

kartik4949:
adds an integration test for raw prompt i.e list of dicts.

def prompt():
    return [{'role': 'user', 'content': ...}]


RyanMarten (approved):
Thank you!

Issue #484, opened 2025-02-10
Request: Queries Per Second rate limiter for Together.AI (fastest R1 provider atm)


alpayariyak:
Hi,

Thank you so much for adding concurrent request rate limiter quickly! 

Could we also add Queries Per Second limiter? Together.AI is the only provider I know that uses it, but it’s currently the fastest for R1, so it would be nice to have :)

![Image](https://github.com/user-attachments/assets/91e8cde2-0d2a-459b-a5c7-2202dfa2e3c8)

RyanMarten:
From the docs https://docs.together.ai/docs/rate-limits
>How are our rate limits implemented?
 Our rate limits are currently measured in **requests per second (RPS)** and **tokens per second (TPS)** for each model type. If you exceed any of the rate limits you will get a 429 error. We show you the values per minute below, as its the industry standard.

| Tier   | Qualification criteria               | RPM   | TPM       |
| ------ | ------------------------------------ | ----- | --------- |
| Free   | User must be in an allowed geography | 60    | 60,000    |
| Tier 1 | Credit card added                    | 600   | 180,000   |
| Tier 2 | $50 paid                             | 1,800 | 250,000   |
| Tier 3 | $100 paid                            | 3,000 | 500,000   |
| Tier 4 | $250 paid                            | 4,500 | 1,000,000 |
| Tier 5 | $1,000 paid                          | 6,000 | 2,000,000 |


> Note: Due to high demand, DeepSeek R1 has a model specific rate limit of 6 RPM for Free Users, and 12 RPM for Build Tier 1 users.

@alpayariyak Are the higher tier rate limits above correct for DeepSeek-R1? I have pasted the rate limit header information that I see below (I am tier 5). 

If this information is per-second, it would suggest
250,000 TPM
1,200 RPM 


Executing
```
 curl -X POST "https://api.together.xyz/v1/chat/completions" \
  -H "Authorization: Bearer $TOGETHER_API_KEY" \
  -H "Content-Type: application/json" \
  -d '{
    "model": "deepseek-ai/DeepSeek-R1",
    "messages": [{"role": "user", "content": "What are some fun things to do in New York?"}]
  }'\
  -i
```

Gives me 
```
HTTP/2 200
...
x-inference-version: v2
x-ratelimit: false
x-ratelimit-limit: 10
x-ratelimit-limit-tokens: 4166
x-ratelimit-remaining: 19
x-ratelimit-remaining-tokens: 4166
[...]

PR #490, opened 2025-02-11
Ref/ratelimit/togetherai

kartik4949:
addresses #484


RyanMarten (approved):
LGTM! Note that we can fix the litellm backend usage of togetherai in a future PR

Issue #485, opened 2025-02-10
Using together through litellm background prints lots of warnings


RyanMarten:
`llm = LLM(model_name="together_ai/deepseek-ai/DeepSeek-R1")`

```
10:06:53 - LiteLLM:WARNING: chat.py:35 - Only some together models support function calling/response_format. Docs - https://docs.together.ai/docs/function-calling
10:06:53 - LiteLLM:WARNING: chat.py:35 - Only some together models support function calling/response_format. Docs - https://docs.together.ai/docs/function-calling
10:06:53 - LiteLLM:WARNING: chat.py:35 - Only some together models support function calling/response_format. Docs - https://docs.together.ai/docs/function-calling
10:06:53 - LiteLLM:WARNING: chat.py:35 - Only some together models support function calling/response_format. Docs - https://docs.together.ai/docs/function-calling
10:06:53 - LiteLLM:WARNING: chat.py:35 - Only some together models support function calling/response_format. Docs - https://docs.together.ai/docs/function-calling
10:06:53 - LiteLLM:WARNING: chat.py:35 - Only some together models support function calling/response_format. Docs - https://docs.together.ai/docs/function-calling
10:06:53 - LiteLLM:WARNING: chat.py:35 - Only some together models support function calling/response_format. Docs - https://docs.together.ai/docs/function-calling
10:06:53 - LiteLLM:WARNING: chat.py:35 - Only some together models support function calling/response_format. Docs - https://docs.together.ai/docs/function-calling
10:06:53 - LiteLLM:WARNING: chat.py:35 - Only some together models support function calling/response_format. Docs - https://docs.together.ai/docs/function-calling
10:06:53 - LiteLLM:WARNING: chat.py:35 - Only some together models support function calling/response_format. Docs - https://docs.together.ai/docs/function-calling
10:06:53 - LiteLLM:WARNING: chat.py:35 - Only some together models support function calling/response_format. Docs - https://docs.together.ai/docs/function-calling
10:06:53 - LiteLLM:WARNING: chat.py:35 - Only some together models support function calling/response_format. Docs - https://docs.together.ai/docs/function-calling
10:06:53 - LiteLLM:WARNING: chat.py:35 - Only some together models support function calling/response_format. Docs - https://docs.together.ai/docs/function-calling
10:06:53 - LiteLLM:WARNING: chat.py:35 - Only some together models support function calling/response_format. Docs - https://docs.together.ai/docs/function-calling
[...]

PR #509, opened 2025-02-17
Update/litellm

kartik4949:
fix https://github.com/bespokelabsai/curator/issues/485


src/bespokelabs/curator/cost.py
```diff
@@ -38,11 +38,11 @@ def cost(self, *, completion_window="*", **kwargs):
         if "completion_response" in kwargs:
             model = kwargs["completion_response"]["model"]
         else:
-            model = kwargs.get("model", None)
+            model = kwargs.pop("model", None)
```
vutrung96, line 41: why pop instead of get?

src/bespokelabs/curator/cost.py
```diff
@@ -38,11 +38,11 @@ def cost(self, *, completion_window="*", **kwargs):
...
 
         cost_to_complete = 0.0
         if model in litellm.model_cost:
-            cost_to_complete = litellm.completion_cost(**kwargs)
+            cost_to_complete = litellm.completion_cost(model=model, **kwargs)
```
vutrung96, line 45: oh so previously we just didn't pass in the model at all?

src/bespokelabs/curator/request_processor/online/litellm_online_request_processor.py
```diff
@@ -220,7 +220,8 @@ def test_call(self):
         )
         # Try the method of caculating cost
         try:
-            litellm.completion_cost(completion_response=completion.model_dump())
+            completion_response = completion.model_dump()
+            litellm.completion_cost(completion_response=completion_response, model=completion_response["model"])
```
vutrung96, line 224: why do we not use _LitellmCostProcessor for this?

src/bespokelabs/curator/cost.py
kartik4949: with new updated litellm version 
modell is required mandatorily outside

PR #516, opened 2025-02-18
Rich hyperlink for curator viewer

### LINKED ISSUE #517: UX: better clickable curator viewer links
Image

if the width of the console is too small, it will break the link into two lines. Which the user can't just command+click onto the link to redirect, or else, it will only copy the https://curator.bespokelabs.ai/datasets/ef3807724cd94a but not the rest 639a821f659ca4e64c, link got cutoff

CharlieJCJ:
https://github.com/user-attachments/assets/7930511c-61c9-4599-b756-5523ece420d5

Resolves #517


kartik4949 (approved):
Thanks!!

src/bespokelabs/curator/client.py
```diff
@@ -56,7 +56,12 @@ def create_session(self, metadata: t.Dict):
 
         if response.status_code == 200:
             self._session = response.json().get("session_id")
-            logger.info("View your data live at: " + f"{PUBLIC_CURATOR_VIEWER_URL}/{self.session}")
+            logger.info(
```
vutrung96, line 59: can you push the change we discussed offline @CharlieJCJ? (make the link part of console)

vutrung96 (changes requested):

src/bespokelabs/curator/client.py
CharlieJCJ: Pushed. 

For this particular logger info, I kept it. 

This info will also be visible from the rich console

vutrung96 (approved):

Issue #508, opened 2025-02-17
Epic: Better Logging.


kartik4949:
- Add a Logger module to implement a custom logger class to define log file, levels, format etc.
- This class instance can be created in each module
- Public api interface to define log level

Example

```python
from bespokelabs.curator import logger
logger.pin()

logger.info("Some pine")
```

The above should create a log file `.log` in `cache` folder which has all the logs with DEBUG level

PR #518, opened 2025-02-19
Feat/logger

kartik4949:
address #508

### DIFF
--- a/src/bespokelabs/curator/log.py
+++ b/src/bespokelabs/curator/log.py
@@ -5,8 +5,6 @@ import os
 from rich.console import Console
 from rich.logging import RichHandler
 
-from bespokelabs.curator.constants import _CURATOR_DEFAULT_CACHE_DIR
-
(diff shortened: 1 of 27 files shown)


src/bespokelabs/curator/log.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def __new__(cls):
+        """Python new method."""
+        if cls._instance is None:
+            cls._instance = super().__new__(cls)
+            cls._instance._init_logger()
+        return cls._instance
+
+    def _init_logger(self):
+        self.logger = logging.getLogger("curator")
```
vutrung96, line 27: should this be bespokelabs.curator?

vutrung96 (approved):
thanks for doing this!

src/bespokelabs/curator/log.py
kartik4949: ah sure

PR #520, opened 2025-02-19
Feat/push to viewer

kartik4949:
Create a utility for pushing hugging face dataset to curator viewer

### DIFF
--- a/src/bespokelabs/curator/utils.py
+++ b/src/bespokelabs/curator/utils.py
@@ -5,7 +5,7 @@ import typing as t
 import uuid
 
 from datasets import Dataset, load_dataset
-from tqdm import tqdm
+from rich.progress import Progress
 
 from bespokelabs.curator import _CONSOLE, constants
 from bespokelabs.curator.client import Client, _SessionStatus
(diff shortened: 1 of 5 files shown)


src/bespokelabs/curator/utils.py
```diff
@@ -0,0 +1,70 @@
...
+        "status": _SessionStatus.STARTED,
+    }
+
+    session_id = client.create_session(metadata, verbose=False)
+    if not client.session:
+        logger.error("Failed to create session.")
+        raise Exception("Failed to create session.")
+
+    view_url = f"{constants.PUBLIC_CURATOR_VIEWER_DATASET_URL}/{session_id}"
+    logger.info(f"View your data live at: {view_url}")
```
vutrung96, line 50: can you reuse the logging logic here for consistency and also to avoid URL cutoff issue that @CharlieJCJ found?

https://github.com/bespokelabsai/curator/commit/dc306aaa2371dc7ace1929abb22dd7298f29bf2c#diff-818ecdf372f8381883569c7bb7ab090e915dba27efbab3c41e3bfd692077c10fR202

src/bespokelabs/curator/utils.py
```diff
@@ -0,0 +1,70 @@
...
+    num_shards = (len(dataset) // chunk_size) + 1
+
+    async def send_responses():
+        async def send_row(idx, row):
+            response_data = {"parsed_response_message": [row]}
+            response_data_json = json.dumps(response_data)
+            await client.stream_response(response_data_json, idx)
+
+        async def process_shard(shard_idx):
+            shard = dataset.shard(num_shards=num_shards, index=shard_idx)
```
vutrung96, line 60: can this be arbitrarily large and just overwhelm the user's bandwidth? should we set some reasonable max concurrent requests with semaphore?

vutrung96 (changes requested):

src/bespokelabs/curator/utils.py
kartik4949: its controlled via `chunk_size`
implicitly its acting like a semaphore

src/bespokelabs/curator/utils.py
kartik4949: done

src/bespokelabs/curator/utils.py
vutrung96: wait but to send fewer requests we are forced to increase `chunk_size` which can also overwhelm bandwidth no?

src/bespokelabs/curator/utils.py
```diff
@@ -52,22 +53,22 @@ def push_to_viewer(dataset: Dataset | str, hf_params: t.Optional[t.Dict] = None,
         f"[dim]{view_url}[/dim]\n"
     )
     _CONSOLE.print(viewer_text)
-    num_shards = (len(dataset) // chunk_size) + 1
+    semaphore = asyncio.Semaphore(max_concurrent_requests)
+
+    progress_bar = tqdm(total=len(dataset), desc="Uploading dataset rows", position=0, leave=True)
```
vutrung96, line 58: can we use a rich progress bar: https://github.com/bespokelabsai/curator/blob/74d11a86347b1339137e5cd992db8a460a3440d4/src/bespokelabs/curator/status_tracker/online_status_tracker.py#L114C1-L123C1? 

looks much nicer and we want to have things look nice everywhere in curator.

src/bespokelabs/curator/utils.py
kartik4949: I moved to semaphore restricted version.
its simpler 
thanks

src/bespokelabs/curator/utils.py
kartik4949: yes, good catch

kartik4949: done

vutrung96 (approved):

PR #533, opened 2025-02-24
Feat/resume/hosted curator viewer

kartik4949:
- Support resume for curator hosted viewer

### DIFF
--- a/src/bespokelabs/curator/db.py
+++ b/src/bespokelabs/curator/db.py
@@ -45,7 +45,7 @@ class MetadataDB:
             "batch_mode",
             "created_time",
             "last_edited_time",
-            "hosted_viewer",
+            "is_hosted_viewer_synced",
         ]
         current_info = self._get_current_schema()
         current_columns = [col[1] for col in current_info]  # col[1] = column name
@@ -89,7 +89,7 @@ class MetadataDB:
                     batch_mode BOOLEAN,
                     created_time TEXT,
                     last_edited_time TEXT,
-                    hosted_viewer BOOLEAN
+                    is_hosted_viewer_synced BOOLEAN
                 )
                 """
             )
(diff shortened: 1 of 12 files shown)


vutrung96 (commented):
will this support the following scenario (I don't think so):

1/ responses were fully generated with viewer disabled
2/ user enables viewer, then runs an identical curator run and gets a cache hit

src/bespokelabs/curator/db.py
```diff
@@ -37,13 +37,15 @@ def validate_schema(self):
...
             "run_hash",
+            "session_id",
             "dataset_hash",
             "prompt_func",
             "model_name",
             "response_format",
             "batch_mode",
             "created_time",
             "last_edited_time",
+            "hosted_viewer",
```
vutrung96, line 48: maybe change this to is_hosted_viewer_synced to be more descriptive?

src/bespokelabs/curator/llm/llm.py
```diff
@@ -242,6 +249,17 @@ def __call__(
                 prompt_formatter=self.prompt_formatter,
             )
 
+        if existing_session_id is not None and existing_hosted_viewer is False:
+            if self._request_processor.viewer_client.hosted:
+                logger.info(
+                    f"Existing metadata found for run hash {metadata_dict['run_hash']} without HOSTED_VIEWER flag.\n"
```
vutrung96, line 255: "Existing metadata found for run hash {metadata_dict['run_hash']} without HOSTED_VIEWER flag." is a bit unclear for me.

"There was a previous run with the same run hash ({metadata_dict['run_hash']}) without the HOSTED_CURATOR_VIEWER flag enabled, and HOSTED_CURATOR_VIEWER flag is enabled for this run. This means that the Curator Viewer is potentially inconsistent with local data. Pushing the full dataset to the Curator Viewer to ensure full consistency."

vutrung96 (approved):

src/bespokelabs/curator/llm/llm.py
kartik4949: done

src/bespokelabs/curator/db.py
kartik4949: done

PR #555, opened 2025-02-28
fix: publish package install command

RyanMarten:
To ensure all dependencies are downloaded for testing (needs to include extras and dev)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ clean:
 
 install:
 	@echo "Installing dependencies..."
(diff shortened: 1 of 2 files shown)


kartik4949 (approved):
Thanks for adding this

kartik4949 (changes requested):
poetry install --extras "code_execution vllm" --with dev
add this in makefile as well?

kartik4949 (approved):

PR #574, opened 2025-03-03
fix: make process_response mandatory in create_dataset files

### LINKED ISSUE #519: Bug: Changing "parse" does not result in newly parsed dataset
I found that after getting the cached dataset, if I change the `parse` method in my defined class of curator.LLM, then if I run `result = llm(ds)` again `result` will not be parsed in the new way.

Minimal example:

```
import os
import getpass
os.environ["OPENAI_API_KEY"] = getpass.getpass()

from typing import Dict
from pydantic import BaseModel, Field
from bespokelabs import curator
from datasets import Dataset

class Poem(BaseModel):
    title: str = Field(description="The title of the poem.")
    poem: str = Field(description="The content of the poem.")

class Poet(curator.LLM):
    response_format = Poem

    def prompt(self, input: Dict) -> str:
        return f"Write two poems about {input['topic']}."

    def parse(self, input: Dict, response: Poem) -> Dict:
        return [{"title": response.title, "poem": response.poem}]

poet = Poet(model_name="gpt-4o-mini")
topics = Dataset.from_dict({'topic': ['Dreams of a Robot']})

poems = poet(topics)
print(poems.to_pandas())

class Poet(curator.LLM):
    response_format = Poem

    def prompt(self, input: Dict) -> str:
        return f"Write two poems about {input['topic']}."

    def parse(self, input: Dict, response: Poem) -> Dict:
        return [{"title_xxxxxxx": response.title, "poem": response.poem}]

poet = Poet(model_name="gpt-4o-mini")
topics = Dataset.from_dict({'topic': ['Dreams of a Robot']})

poems = poet(topics)
print(poems.to_pandas())

```

kartik4949:
fix #519

### DIFF
--- a/src/bespokelabs/curator/request_processor/base_request_processor.py
+++ b/src/bespokelabs/curator/request_processor/base_request_processor.py
@@ -412,8 +412,10 @@ class BaseRequestProcessor(ABC):
                                 error_sample.append(str(response.response_errors))
                             continue
 
-                        if response.parsed_response_message is None:
-                            response.parsed_response_message = self._process_response(response)
+                        # TODO: Find a way to not process responses that have already been processed
+                        # We cannot just check if parsed_response_message is not None because it could be from cached previous run
+                        # response.
+                        response.parsed_response_message = self._process_response(response)
 
                         for row in response.parsed_response_message:
                             if isinstance(row, BaseModel):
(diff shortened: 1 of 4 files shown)


vutrung96 (approved):
thank you for fixing this @kartik4949 !

PR #575, opened 2025-03-04
fix: use model name from config in  cost processor

### LINKED ISSUE #569: [Cost] Custom input/output/etc cost that user can specify from our curator interface
one way to do it is through backend_params, or we can create a separate field for specifying cost-related config.

kartik4949:
fix #569


CharlieJCJ:
Testing on examples, the previous poem one passed https://github.com/bespokelabsai/curator/issues/569#issuecomment-2695446181
image

But now observed gemini from litellm would give this warning if I manual set input output token costs. 
image

kartik4949:
@CharlieJCJ Thanks a lot for doing thing, good catch

The litellm provider was hardcoded

CharlieJCJ (approved):
LGTM, thanks!

PR #582, opened 2025-03-07
feat: extract mime_type in Image types

kartik4949:
- We were not extracting mime_type from image url

### DIFF
--- a/src/bespokelabs/curator/types/prompt.py
+++ b/src/bespokelabs/curator/types/prompt.py
@@ -14,6 +14,7 @@ class BaseType(BaseModel):
 
     url: str = Field("", description="The URL of the file.")
     type: t.ClassVar[str] = Field(..., description="The type of the multimodal prompt.")
+    mime_type: str | None = Field(None, description="The MIME type of the file.")
 
     @staticmethod
     def _is_local_uri(path):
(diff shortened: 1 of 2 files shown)


src/bespokelabs/curator/types/prompt.py
```diff
@@ -80,11 +82,16 @@ def check_url_or_content(self):  # noqa: N805
             raise ValueError("Only one of 'url' or 'content' must be provided.")
         return self
 
+    def model_post_init(self, __context):
+        """Run after initialization."""
+        if self.mime_type is None and self.url:
+            mime_type, _ = mimetypes.guess_type(self.url)
+            self.mime_type = mime_type.lower() if mime_type else "image/png"
```
CharlieJCJ, line 89: Minor concern: If the MIME type detection logic fails (e.g., if `mimetypes.guess_type` does not return a valid type), it default to "image/png" incorrectly, will that lead to issues? We should probably warn users when this happens.

CharlieJCJ (approved):

src/bespokelabs/curator/types/prompt.py
kartik4949: sure, good point

PR #583, opened 2025-03-07
feat: add param to disable metadata db

shreyaspimpalgaonkar:
Running multiple CodeExecutor instances at once errors out writes to metadata DB because SqLite implementation doesn't support concurrency. 

Added a param in `backend_params` to disable writes to the db. Example usage: 

```python
hello_executor = HelloExecutor(backend_params={'use_metadata_db': False})
```

### DIFF
--- a/src/bespokelabs/curator/code_executor/types.py
+++ b/src/bespokelabs/curator/code_executor/types.py
@@ -64,6 +64,7 @@ class CodeExecutionResponse(BaseModel):
 class CodeExecutionBackendConfig(BaseModel):
     """Configuration for the code execution backend."""
 
+    use_metadata_db: bool = True
     max_requests_per_minute: int = 10000
     max_retries: int = 3
     seconds_to_pause_on_rate_limit: int = 10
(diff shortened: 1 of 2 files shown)


src/bespokelabs/curator/code_executor/types.py
```diff
@@ -64,6 +64,7 @@ class CodeExecutionResponse(BaseModel):
 class CodeExecutionBackendConfig(BaseModel):
     """Configuration for the code execution backend."""
 
+    use_metadata_db: bool = True
```
kartik4949, line 67: its more like exposing internal detail to user?

we should rather disable it internally if inside a concurrent environment i.e 
check number of instance create `CodeExecutor..`  and if more than 1 make it disable internally?

shreyaspimpalgaonkar: Agree, we can clean it up if more users ask. This quick fix is ok for now I think.

kartik4949 (approved):
LGTM

PR #584, opened 2025-03-11
feat: Implement Mistral batch request processor

Saharsh1005:


src/bespokelabs/curator/request_processor/batch/mistral_batch_request_processor.py
```diff
@@ -149,9 +150,9 @@ def parse_api_specific_batch_object(self, mistral_batch_object: BatchJobOut, req
             created_at=mistral_batch_object.created_at,
             finished_at=mistral_batch_object.completed_at,
             status=status,
-            api_key_suffix=self.client.api_key[-4:],
+            api_key_suffix="HAuW",
```
RyanMarten, line 153: this shouldn't be hardcoded

src/bespokelabs/curator/request_processor/batch/mistral_batch_request_processor.py
```diff
@@ -0,0 +1,411 @@
...
+    @property
+    def backend(self):
+        """Backend property."""
+        return "mistral"
+
+    @property
+    def max_requests_per_batch(self) -> int:
+        """The maximum number of requests that can be processed in a batch."""
+        return 1000000  # 1 million
+
```
RyanMarten, line 81: If these maximums are not documented anywhere - let's add that to the comment (that these are arbitrary). 
If the maximums are documented somewhere - then let's put that link in the comment

src/bespokelabs/curator/request_processor/batch/mistral_batch_request_processor.py
```diff
@@ -0,0 +1,411 @@
...
+        """
+        return GenericBatchRequestCounts(
+            failed=mistral_batch_object.failed_requests,
+            succeeded=mistral_batch_object.succeeded_requests,
+            total=mistral_batch_object.total_requests,
+            raw_request_counts_object=mistral_batch_object.model_dump(),
+        )
+
+    def parse_api_specific_batch_object(self, mistral_batch_object: BatchJobOut, request_file: str | None = None) -> GenericBatch:
+        """Convert a Mistral batch object to generic format.
```
RyanMarten, line 115: Thanks for the detailed doc string! :D

src/bespokelabs/curator/request_processor/batch/mistral_batch_request_processor.py
```diff
@@ -0,0 +1,411 @@
...
+        Args:
+            mistral_batch_object: Mistral's Batch object.
+            request_file: Optional path to the request file.
+
+        Returns:
+            GenericBatch: Standardized batch object.
+
+        Reference: Mistral client-python: https://github.com/mistralai/client-python/blob/main/docs/models/batchjobout.md
+        """
+        if MistralProgressStates.has_value(mistral_batch_object.status):
```
RyanMarten, line 140: Nice, I like this. We can also refactor the other batch request processors to use this clean approach (in a separate PR).

src/bespokelabs/curator/request_processor/batch/mistral_batch_request_processor.py
```diff
@@ -0,0 +1,411 @@
...
+            finished_at=batch.finished_at,
+            token_usage=token_usage,
+            response_cost=cost,
+        )
+        return generic_response
+
+    def create_api_specific_request_batch(self, generic_request: GenericRequest) -> dict:
+        """Creates an API-specific request body from a generic request body.
+
+        Transforms a GenericRequest into the format expected by MIstral's batch API.
```
RyanMarten, line 212: "MIstral" - typo

src/bespokelabs/curator/request_processor/batch/mistral_batch_request_processor.py
```diff
@@ -0,0 +1,411 @@
...
+    async def upload_batch_file(self, file_content: bytes) -> UploadFileOutTypedDict:
+        """Uploads a batch file to Mistral and waits until ready.
+
+        Args:
+            file_content (bytes): The encoded file content to upload
+
+        Returns:
+            UploadFileOutTypedDict: The uploaded file object from Mistral
+        """
+        with tempfile.NamedTemporaryFile(delete=True) as temp_file:
```
RyanMarten, line 250: I don't think you need to create the temporary file at all actually. Since self.client.files.upload takes "content" - as you are passing it directly in - you don't need to have a local file saved. In fact, here the temporary file is deleted as the with block exits before the upload. 

The file name seems to just be an identifier for the upload (see https://console.mistral.ai/build/datasets). 
This can be random (e.g. similar to what you are doing here with the temp name). Or it can be something like "curator_cache-fingerprint_requests_i"

src/bespokelabs/curator/request_processor/batch/mistral_batch_request_processor.py
```diff
@@ -0,0 +1,411 @@
...
+    def _download_and_load_output(self, output_file: httpx.Request) -> list[dict]:
+        """Download the output file to a temporary file and load it as a list of dictionaries.
+
+        Args:
+            output_file: The file to download and parse.
+
+        Returns:
+            list[dict] | None: The parsed JSON objects if successful, None if failed.
+        """
+        import json
```
RyanMarten, line 338: move import to the top of the file

src/bespokelabs/curator/request_processor/batch/mistral_batch_request_processor.py
```diff
@@ -0,0 +1,411 @@
...
+
+        with tempfile.NamedTemporaryFile(delete=True, mode="w+", encoding="utf-8") as temp_file:
```
RyanMarten, line 340: Do you even need to materialize to a file here? You can load the JSON object directly from the file content.

RyanMarten (changes requested):

src/bespokelabs/curator/request_processor/batch/mistral_batch_request_processor.py
Saharsh1005: Alright, I will make this modification.

Saharsh1005:
#### cost calc -- doubt/clarification
`Model Pricing: Per 1M tokens - Input: $0.000 • Output: $0.000` — I'm unsure how this value is generated when running curator. My guess is that `liteLLM`, using its `cost tables`, is unable to determine the costs.

I'll double-check the cost calculations for my test case. If I recall correctly, since token usage was very low, the cost variable held a small value, but the final result was 0. I can test with `request/response` pairs that have higher token usage to verify.

src/bespokelabs/curator/request_processor/_factory.py
```diff
@@ -156,6 +160,20 @@ def create(
             from bespokelabs.curator.request_processor.offline.vllm_offline_request_processor import VLLMOfflineRequestProcessor
 
             _request_processor = VLLMOfflineRequestProcessor(config)
+
+        elif backend == "mistral" and not batch:
+            config.api_key = config.api_key or os.getenv("MISTRAL_API_KEY")
+            if not config.api_key:
+                raise ValueError("MISTRAL_API_KEY is not set")
+            raise ValueError("Only batch mode is supported with Mistral backend")
```
kartik4949, line 168: good

src/bespokelabs/curator/request_processor/batch/mistral_batch_request_processor.py
```diff
@@ -0,0 +1,411 @@
...
+
+    Reference for Mistral status: https://github.com/mistralai/client-python/blob/main/docs/models/batchjobstatus.md.
+    """
+
+    SUCCESS = "SUCCESS"
+    FAILED = "FAILED"
+    TIMEOUT_EXCEEDED = "TIMEOUT_EXCEEDED"
+    CANCELLED = "CANCELLED"
+
+    @classmethod
```
kartik4949, line 50: Create a BaseState and add this `has_value` method in it.
and subclass these two classes.

Later we will implement other backend in similar fashion as well

src/bespokelabs/curator/request_processor/batch/mistral_batch_request_processor.py
kartik4949: you can add a CONSTANT at top may be

src/bespokelabs/curator/request_processor/batch/mistral_batch_request_processor.py
```diff
@@ -0,0 +1,411 @@
...
+            cost = None
+        else:
+            response_message = raw_response["response"]["body"]["choices"][0]["message"]["content"]
+            response_errors = None
+            token_usage = _TokenUsage(
+                input=int(raw_response["response"]["body"]["usage"]["prompt_tokens"]),
+                output=int(raw_response["response"]["body"]["usage"]["completion_tokens"]),
+                total=int(raw_response["response"]["body"]["usage"]["total_tokens"]),
+            )
+            cost = self._cost_processor.cost(model="mistral/" + self.config.model, prompt=str(generic_request.messages), completion=response_message)
```
kartik4949, line 194: "mistral/" prefix is for litellm cost processor I guess.

are you sure, 
"mistral/{self.config.model}" works for all mistral models?

src/bespokelabs/curator/request_processor/batch/mistral_batch_request_processor.py
```diff
@@ -0,0 +1,411 @@
...
+        Returns:
+            UploadFileOutTypedDict: The uploaded file object from Mistral
+        """
+        with tempfile.NamedTemporaryFile(delete=True) as temp_file:
+            temp_file.write(file_content)
+            temp_file.flush()
+
+        batch_file_upload = self.client.files.upload(
+            file={
+                "file_name": temp_file.name,
```
kartik4949, line 256: why can't we upload request_file? like openai?

src/bespokelabs/curator/request_processor/batch/mistral_batch_request_processor.py
```diff
@@ -0,0 +1,411 @@
...
+
+        with tempfile.NamedTemporaryFile(delete=True, mode="w+", encoding="utf-8") as temp_file:
+            for chunk in output_file.stream:
+                temp_file.write(chunk.decode("utf-8"))
+
+            temp_file.flush()
+            temp_file.seek(0)
+
+            raw_content = temp_file.read()
+            try:
```
kartik4949, line 348: why is this complicated?
can't we convert output_file.stream -> parsed_objects?

kartik4949: you can delimit the stream with \n' ?

src/bespokelabs/curator/request_processor/batch/mistral_batch_request_processor.py
RyanMarten: Store a self.api_key in the __init__ and use the [:-4] suffix here. This should also be fixed in the `gemini_batch_request_processor.py` too @kartik4949

tests/integrations/test_mistral_batch.py
```diff
@@ -0,0 +1,39 @@
...
+    """Generate a batch of 4 test prompts to test batch processing."""
+    return [
+        "Describe the impact of quantum computing on AI in 2 sentences.",
+        "Give  a creative business idea for a startup in 2025 in 1 sentence.",
+        "What is 2*10?",
+        "Who is the current prime-minister of India?",
+    ]
+
+
+@pytest.mark.integration
```
kartik4949, line 19: The tests should move to test_all.py
check how vcr works.

append the _BATCH_BACKENDS at the top of the test_all.py 
if you are running the tests 1st time.

do this.

`VCR_MODE=all poetry run pytest ...`
VCR_MODEL = all means all http interactions will be recorded into a file and later we can replay them without actually network calls.

kartik4949 (changes requested):
Great work.
Please do some refactors and correct testing.
THanks

kartik4949:
> #### cost calc -- doubt/clarification
> `Model Pricing: Per 1M tokens - Input: $0.000 • Output: $0.000` — I'm unsure how this value is generated when running curator. My guess is that `liteLLM`, using its `cost tables`, is unable to determine the costs for `mistral/mistral-tiny`.
> 
> I'll double-check the cost calculations for my test case. If I recall correctly, since token usage was very low, the cost variable held a small value, but the final result was 0. I can test with `request/response` pairs that have higher token usage to verify.

Hey could you check 
`base_batch_request_processor` and see around L313
how 
           ` self.tracker.input_cost_per_million = (model_cost[self.prompt_formatter.model_name]["input_cost_per_token"] * 1_000_000) * 0.5` is done.

`model_cost` is a llitellm thing, and `self.config.model_name` does not have `mistral/` prefix so you need to check the other logic i.e `external_model_cost`

src/bespokelabs/curator/request_processor/batch/mistral_batch_request_processor.py
Saharsh1005: @RyanMarten Thank you for your detailed review. I have made the modifications as requested.

src/bespokelabs/curator/request_processor/batch/mistral_batch_request_processor.py
Saharsh1005: `file` dictionary that we pass to `Mistral client` requires filed `fileName`.
I have modified my implementation as per @RyanMarten's suggestion

src/bespokelabs/curator/request_processor/batch/mistral_batch_request_processor.py
Saharsh1005: Rectified as per your and Ryan's suggestions. Initially, I was unnecessarily creating temporary files along the way.

src/bespokelabs/curator/request_processor/batch/mistral_batch_request_processor.py
Saharsh1005: Implemented https://github.com/bespokelabsai/curator/pull/584/commits/37e9666bb2f1ff290496aa45a512e88b379b9e50

src/bespokelabs/curator/request_processor/batch/mistral_batch_request_processor.py
Saharsh1005: Initially, I had referred to this JSON document: [LiteLLM Model Prices and Context Window](https://github.com/BerriAI/litellm/blob/main/model_prices_and_context_window.json) and wrote the code accordingly.
Line 233: Added the `"mistral/"` prefix: `litellm.get_max_tokens("mistral/" + self.config.model)`
Line 194: Updated the model parameter with the `"mistral/"` prefix:
```
self._cost_processor.cost(
    model="mistral/" + self.config.model,
    prompt=str(generic_request.messages),
    completion=response_message
)
```
However, upon a closer look at the JSON document, I have found that it also includes Mistral models with an `"cloudflare/@cf" "azureai/" "cloudfare/" "snowflake" "azure"` provider prefix, which our current logic does not handle. 

We need to account for this to avoid potential mismatches.

### Possible solution:
Should we a) store these hardcoded OR b) write a function (eg: `_get_model_prefix`) which dynamically fetches the file from the LiteLLM GitHub repository and looks for all `mistral` providers?

cc: @kartik4949

src/bespokelabs/curator/request_processor/batch/mistral_batch_request_processor.py
Saharsh1005: `fileName` is a required field which needs to be passed in the `file` dictionary as per MistralAPI.
I have simplified this as per Ryans suggestion above in my latest commit https://github.com/bespokelabsai/curator/pull/584/commits/caa7a70 (line:254)

src/bespokelabs/curator/request_processor/batch/mistral_batch_request_processor.py
Saharsh1005: rectified the `lazy` import.

src/bespokelabs/curator/request_processor/batch/mistral_batch_request_processor.py
Saharsh1005: rectified.

RyanMarten (approved):
Thanks for making the changes @Saharsh1005! Great PR :D 
Once you fix issues with cost and testing that @kartik4949 identified, we can merge it in! 🥳

src/bespokelabs/curator/request_processor/batch/mistral_batch_request_processor.py
```diff
@@ -0,0 +1,383 @@
...
+
+        Args:
+            file_content (bytes): The encoded file content to upload
+
+        Returns:
+            UploadFileOutTypedDict: The uploaded file object from Mistral
+        """
+        batch_file_upload = self.client.files.upload(
+            file={
+                "file_name": f"curator_cache-fingerprint_requests_{uuid.uuid4().hex}",  # Random placeholder name
```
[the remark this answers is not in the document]
Saharsh1005: rectified.

tests/integrations/test_mistral_batch.py
Saharsh1005: This is a great approach for offline testing and replicating API-based testing!
I will look into this in-depth whenever I get a chance.

Meanwhile I tried to transfer `mistral_batch_test` to `test_all`,
But face this issue:

```
E           mistralai.models.sdkerror.SDKError: API error occurred: Status 401
E           {
E             "message":"Unauthorized",
E             "request_id":"da9e8d73f25f0336fa85b02444e4042f"
E           }
```

command: `pytest tests/integrations/test_all.py -k mistral -s`

src/bespokelabs/curator/request_processor/batch/mistral_batch_request_processor.py
Saharsh1005: `cost lookup` impllementation has been rectified in the lastest commit!
Thanks @kartik4949  :)!

tests/integrations/test_mistral_batch.py
Saharsh1005: implemented as per our discussion last week.

src/bespokelabs/curator/request_processor/batch/mistral_batch_request_processor.py
kartik4949: Thanks

Saharsh1005:
> @Saharsh1005 could u poetry lock it? and also remove the test from test_all.py and mark test_mistral_batch.py as skipped.
> 
> There is a weird locking behaviour of mistral test when used with vcr

HI @kartik4949 
I have rectified the same in the latest commit!

tests/integrations/test_all.py
```diff
@@ -27,7 +27,7 @@ def _hash_string(input_string):
 _ONLINE_BACKENDS = [{"integration": backend} for backend in {"openai", "litellm"}]
 _ONLINE_CONCURRENT_ONLY_BACKENDS = [{"integration": backend} for backend in {"litellm/deepinfra"}]
 _FAILED_BATCH_BACKENDS = [{"integration": backend, "cached_working_dir": True} for backend in {"anthropic", "openai"}]
-_BATCH_BACKENDS = [{"integration": backend} for backend in {"anthropic", "openai"}]
+_BATCH_BACKENDS = [{"integration": backend} for backend in {"anthropic", "openai", "mistral"}]
```
kartik4949, line 30: remove this mistral

kartik4949:
@Saharsh1005 I have created a small pr on your repo, either accept it or do the same changes here
Thanks

Saharsh1005:
> @Saharsh1005 I have created a small pr on your repo, either accept it or do the same changes here Thanks

Hi @kartik4949 Done, I have removed `mistral` occurrences from`test_all` and `conftest`.

tests/conftest.py
```diff
@@ -15,6 +15,7 @@
     "deepinfra": "DEEPINFRA_API_KEY",
     "vllm": "VLLM_API_KEY",
     "gemini": "GEMINI_API_KEY",
+    "mistral": "MISTRAL_API_KEY",
```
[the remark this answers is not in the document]
Saharsh1005: resolved.

tests/integrations/test_all.py
Saharsh1005: resolved.

kartik4949:
@Saharsh1005 I will add vcr test in separate pr

RyanMarten:
@Saharsh1005 

Can you merge in @kartik4949 changes please? 
https://github.com/Saharsh1005/bespokeai-curator/pull/1/files

Saharsh1005:
Changes have been merged.

Best,
Saharsh

kartik4949 (approved):

RyanMarten:
Congrats on your first contribution!!!! @Saharsh1005 ❤️🎉🍾🥳

PR #591, opened 2025-03-14
fix: use semaphore to gate create new rows to prevent OOMs

vutrung96:
previously we were only using semaphore to gate sending the response, but not the creation of the task itself. this essentially means we load the entire dataset in memory, since we will create one task in-memory for each row. for large datasets, this causes OOMs.

we need to gate the task creation with semaphore.


kartik4949 (approved):
Thanks

PR #599, opened 2025-03-23
Fix/live colab display error

CharlieJCJ:
Previously in colab, when there's keyboard interrupt, the second time running curator, would get this error message
image

The error LiveError: Only one live display may be active at once occurs because trying to create a second Rich Live display while one is already active. This typically happens when a previous run wasn't properly terminated, which is common after keyboard interruption.

Fixing this behavior. 

Also in this PR, lint for online status tracker.


shreyaspimpalgaonkar (approved):
Looks good

PR #603, opened 2025-03-25
Update client.py

sharatsc:
Remove posthog API key from repo.
Should be cycled.


shreyaspimpalgaonkar:
Hey Sharat,  thanks for your contribution! This is a post-only public key pointing to bespokelabs' posthog which we use for basic analytics on the usage of curator. We should keep it public as it won't be available in users' environments. As the current implementation is okay, I'll go ahead and close this PR. Thanks!

PR #604, opened 2025-03-26
If the input is a string or a list of strings for prompt(), maintain the same for parse()

madiator:
When a user passes a string or a list of strings, we create a dataset object with key __internal_prompt.
We remove this when passing to prompt(), but keep it when passing to parse(). So fix this step.


kartik4949 (approved):
LGTM!

PR #606, opened 2025-03-27
feat: add authentication flow in curator client

kartik4949:
- Adds authenticated communication with curate viewer via api key


vutrung96 (approved):
thank you!

Issue #611, opened 2025-04-01
parse errors happening in gemini batch


suhani221:
"/shared/home/suhani/miniconda3/envs/my_env/lib/python3.12/site-packages/bespokelabs/curator/request_processor/batch/gemini_batch_request_proces
sor.py", line 297, in parse_api_specific_response
    response_message_raw = response_body["candidates"][0]["content"]["parts"][0]["text"]
               

KeyError: 'parts', indicates that the code is trying to access a key called parts in the dictionary, but it doesn't exist in the response body. This usually happens when the structure of the data being returned by the API is different from what the code expects.

kartik4949:
@suhani221 we are aware of this issue and it seems like it is due to gemini batch api itself returning empty content.

you can confirm this by checking raw outputs stored on your gemini bucket.
Thanks

suhani-sarvam:
What could be the way around this?Entire process comes to halt because of it.

kartik4949:
I have created a pr to handle this situation 
https://github.com/bespokelabsai/curator/pull/612

kartik4949:
@suhani-sarvam could also attach the raw response you got in your google bucket?

suhani-sarvam:
I deleted all the cache files, sorry😕

kartik4949:
@suhani-sarvam its on your google bucket :)

suhani221:
there were 1more than l rows and it created like 12 requests file, so how do I find the exact one?

kartik4949:
@suhani221 no problem, please use the attached branch, it has the fix.

It has to do with Recitation finish reason.
search for  `RECITATION` finish reason in the response file
Thanks

kartik4949:
fixed by #612

PR #612, opened 2025-04-01
feat: fix gemini batch parts key missing

kartik4949:
https://github.com/bespokelabsai/curator/issues/611


vutrung96 (approved):
thank you!

PR #614, opened 2025-04-02
fix: multiple bugs in batch cancellation

vutrung96:
Fixed multiple bugs in batch cancellation and add tests to prevent future regressions.


Anna-Pinewood:
Hi!
I don't think this would be enough, because `cancel_batches` is asynchronious function, and you're calling it without async.

`.venv/lib/python3.11/site-packages/bespokelabs/curator/request_processor/batch/base_batch_request_processor.py`
```
    async def cancel_batches(self):
        """Cancel all currently submitted batches.

        Attempts to cancel all batches that are currently in submitted state.
        Handles cases where no batches are submitted.

        Side Effects:
            - Attempts to cancel all submitted batches concurrently
            - Logs warning if no batches to cancel
            - Updates batch status through cancel_batch calls
        """
        if self.tracker.n_submitted_batches == 0:
            logger.warning("No batches to be cancelled, but cancel_batches=True.")
            return
        tasks = [self.cancel_batch(batch) for batch in self.tracker.submitted_batches.values()]
        await asyncio.gather(*tasks)
```

vutrung96:
Ah good catch :) let me add await and test this properly

src/bespokelabs/curator/request_processor/batch/base_batch_request_processor.py
```diff
@@ -90,6 +90,7 @@ def requests_to_responses(self, generic_request_files: list[str]) -> None:
         self.batch_objects_file = os.path.join(self.working_dir, "batch_objects.jsonl")
 
         self._attempt_loading_batch_status_tracker(generic_request_files)
+        self.set_model_cost()
```
kartik4949, line 93: why?

src/bespokelabs/curator/request_processor/batch/base_batch_request_processor.py
```diff
@@ -747,8 +747,21 @@ async def cancel_batches(self):
             - Logs warning if no batches to cancel
             - Updates batch status through cancel_batch calls
         """
+        self.prompt_formatter = prompt_formatter
+        self.working_dir = working_dir
+        self.total_requests = len(dataset) if dataset is not None else 1
+
+        self.semaphore = asyncio.Semaphore(self.max_concurrent_batch_operations)
+        self._batch_objects_file_lock = asyncio.Lock()
```
kartik4949, line 755: not used..

src/bespokelabs/curator/request_processor/batch/base_batch_request_processor.py
```diff
@@ -747,8 +747,21 @@ async def cancel_batches(self):
...
+        self.batch_objects_file = os.path.join(working_dir, "batch_objects.jsonl")
+
+        generic_request_files = self.create_request_files(dataset)
+
+        self._attempt_loading_batch_status_tracker(generic_request_files)
+
         if self.tracker.n_submitted_batches == 0:
             logger.warning("No batches to be cancelled, but cancel_batches=True.")
             return
+        logger.info("You set cancel_batches=True. Cancelling batches and returning the original dataset without responses.")
```
kartik4949, line 765: do you think we should ask user are you sure? as a input? press Y/N

tests/integrations/test_all.py
```diff
@@ -334,6 +334,37 @@ def _run_loop(func):
...
+            tracker = BatchStatusTracker.model_validate_json(f.read())
+        assert tracker.n_total_requests == 3
+        assert len(tracker.submitted_batches) == 1
+        assert len(tracker.downloaded_batches) == 0
+
+        logger = "bespokelabs.curator.request_processor.batch.base_batch_request_processor"
+        with caplog.at_level(logging.INFO, logger=logger):
+            helper.create_basic(temp_working_dir, mock_dataset, batch=True, backend=backend, batch_cancel=True)
+            resume_msg = "Cancelling batches"
+            assert resume_msg in caplog.text
```
kartik4949, line 365: can we check if the batch_cancel yaml actually hits cancel request

my point is to also test if the cancellation actually happens

kartik4949 (commented):
thanks for pr!
almost ready

src/bespokelabs/curator/llm/llm.py
```diff
@@ -191,11 +197,24 @@ def __call__(
...
-        dataset_hash = dataset._fingerprint if dataset is not None else xxh64("").hexdigest()
-
         disable_cache = os.getenv("CURATOR_DISABLE_CACHE", "").lower() in ["true", "1"]
         fingerprint = self._hash_fingerprint(dataset_hash, disable_cache)
 
+        run_cache_dir = os.path.join(curator_cache_dir, fingerprint)
+        os.makedirs(run_cache_dir, exist_ok=True)
+        add_file_handler(run_cache_dir)
+
+        if batch_cancel:
```
kartik4949, line 207: if batch_cancel and not self.batch_mode:

src/bespokelabs/curator/request_processor/batch/base_batch_request_processor.py
```diff
@@ -736,19 +736,44 @@ async def download_batch_to_response_file(self, batch: GenericBatch) -> str | No
...
         Side Effects:
             - Attempts to cancel all submitted batches concurrently
             - Logs warning if no batches to cancel
             - Updates batch status through cancel_batch calls
         """
+        self.prompt_formatter = prompt_formatter
+        self.working_dir = working_dir
+        self.total_requests = len(dataset) if dataset is not None else 1
+
+        self.semaphore = asyncio.Semaphore(self.max_concurrent_batch_operations)
```
kartik4949, line 760: semaphore is used?

kartik4949 (commented):
few more changes

tests/integrations/test_all.py
vutrung96: anthropic: https://github.com/bespokelabsai/curator/blob/a0681ea627b7df0d32a45a67431484f9f678080d/tests/integrations/anthropic/fixtures/batch_cancel.yaml#L740
openai: https://github.com/bespokelabsai/curator/blob/a0681ea627b7df0d32a45a67431484f9f678080d/tests/integrations/openai/fixtures/batch_cancel.yaml#L548

src/bespokelabs/curator/request_processor/batch/base_batch_request_processor.py
vutrung96: yea it's use for cancellation

kartik4949 (approved):

tests/integrations/test_all.py
vutrung96: added cassette asserts to make sure all captured requests are called (including the /cancel requests)

src/bespokelabs/curator/llm/llm.py
vutrung96: done

PR #621, opened 2025-04-07
fix: add finish reason in gemini batch

kartik4949:


src/bespokelabs/curator/request_processor/batch/gemini_batch_request_processor.py
```diff
@@ -299,20 +300,30 @@ def parse_api_specific_response(
...
                     response_message_raw = response_message_raw["parts"][0]["text"]
-                else:
+                elif "citationMetadata" in candidate:
                     logger.warning("Model returned a citation response, serialize it into a string!")
-                    response_message_raw = json.dumps(response_body["candidates"][0]["citationMetadata"])
+                    response_message_raw = json.dumps(candidate["citationMetadata"])
+                else:
+                    response_message_raw = response_message_raw or ""
+            else:
+                finish_reason = "PROHIBITED_CONTENT"
```
vutrung96, line 321: is this undocumented behavior? how can we know that it's PROHIBITED_CONTENT if there's no candidate? from https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/inference, i only see this:

```
finishReason: The reason why the model stopped generating tokens. If empty, the model has not stopped generating the tokens. Because the response uses the prompt for context, it's not possible to change the behavior of how the model stops generating tokens.
FINISH_REASON_STOP: Natural stop point of the model or provided stop sequence.
FINISH_REASON_MAX_TOKENS: The maximum number of tokens as specified in the request was reached.
FINISH_REASON_SAFETY: Token generation was stopped because the response was flagged for safety reasons. Note that Candidate.content is empty if content filters block the output.
FINISH_REASON_RECITATION: The token generation was stopped because the response was flagged for unauthorized citations.
FINISH_REASON_BLOCKLIST: Token generation was stopped because the response includes blocked terms.
FINISH_REASON_PROHIBITED_CONTENT: Token generation was stopped because the response was flagged for prohibited content, such as child sexual abuse material (CSAM).
FINISH_REASON_SPII: Token generation was stopped because the response was flagged for sensitive personally identifiable information (SPII).
FINISH_REASON_MALFORMED_FUNCTION_CALL: Candidates were blocked because of malformed and unparsable function call.
FINISH_REASON_OTHER: All other reasons that stopped the token
FINISH_REASON_UNSPECIFIED: The finish reason is unspecified.
```

kartik4949: @vutrung96 Sarvam noted this :)

vutrung96 (approved):

PR #626, opened 2025-04-11
update metadata schema with cost

kartik4949:
PR adds cost management information in session metadata table.


vutrung96:
I'll merge now, we can add tests for these later.

PR #629, opened 2025-04-12
fix/curator CLI display - rpm display and frequent tqdm update for tqdm for online

CharlieJCJ:
Online status tracker fixes.

Changes:
- The tqdm description updates more frequently
- Negative RPM fixed, now it’s req/min and res/min

image
image


kartik4949 (approved):
LGTM!

PR #630, opened 2025-04-12
Fix/deepseek/rpm

kartik4949:
- There is a nasty bug in concurrent strategy where we were not releasing semaphore for success strategy
- max_tpm should be `None` for deepseek by default.


RyanMarten (approved):
Looks good!

PR #638, opened 2025-04-15
cost streaming in batch mode and delayed streaming support

kartik4949:
This pr streams cost information in batch mode to viewer and also adds support for caching updates for every 10 seconds


vutrung96:
could you fix tests?

PR #639, opened 2025-04-15
feat: fix batch mode structure outputs bug

kartik4949:
There is a bug in batch mode for structured outputs

The GenericResponse file with `BaseModel` does not get serialized when `model_dump` is called.


joao-krauchenco-mottu (approved):

CharlieJCJ (approved):
tested on poem.py with `batch=True`, LGTM

image

PR #643, opened 2025-04-16
feat: add response object in curator

kartik4949:
This pr added a feature which enables curator to return a `CuratorResponse` object, which encapsulates all the information about the run as a python object
like, token usage, requests stats, performance stats, failed_requests etc


vutrung96 (commented):
logic looks good.

since this is a breaking change, i think we need to notify users somehow. could we print a warning when the user is trying to access a non-existent attribute, e.g. when they are trying to use it an hf dataset object, that our API changed? 

we should also update the doc.

vutrung96 (approved):
thank you!

PR #661, opened 2025-05-06
feat: add example of prescription extraction

kartik4949:
This pr adds an example which extracts medical prescriptions into a structured format


shreyaspimpalgaonkar:
A readme with some text, screenshot of input and output would be nice to have.

PR #675, opened 2025-05-24
add default app id parameter for curator llm

arnavj053:
when use curator llm to create your own dataset, it will load with default app id parameter that was provided.

### DIFF
--- a/src/bespokelabs/curator/llm/llm.py
+++ b/src/bespokelabs/curator/llm/llm.py
@@ -78,6 +78,7 @@ class LLM:
         generation_params: dict | None = None,
         backend_params: BackendParamsType | None = None,
         system_prompt: str | None = None,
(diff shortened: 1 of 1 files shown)


src/bespokelabs/curator/llm/llm.py
```diff
@@ -261,6 +265,7 @@ def __call__(
             "response_format": (str(self.prompt_formatter.response_format.model_json_schema()) if self.prompt_formatter.response_format else "text"),
             "run_hash": fingerprint,
             "batch_mode": self.batch_mode,
+            "default_app_id": self.default_app_id,
```
shreyaspimpalgaonkar, line 268: only send this if it isn't NULL

shreyaspimpalgaonkar (approved):

Issue #676, opened 2025-05-28
Support for structured output with o3


Shanay-Mehta:
Currently, curator does not support structured output with o3. However, openAI does support it. Can we add the support in curator as well?

PR #678, opened 2025-05-29
ref: added o3 support for structured outputs

kartik4949:
fix https://github.com/bespokelabsai/curator/issues/676


vutrung96 (commented):
does o4 support structured output? should we also add those models?

kartik4949:
> does o4 support structured output? should we also add those models?

Ill address this in separate pr, 
this is fixed by mahesh

PR #690, opened 2025-06-13
Fix Multimodal Gemini Batch Request Creation

alexisdrakopoulos:
## What does this PR do?

This pull request adds support for multimodal inputs (specifically text and images) to the `GeminiBatchRequestProcessor`. The existing implementation only handled text-based requests, causing errors when a `curator.types.Image` was included in the prompt.

This change allows users to leverage Gemini's powerful vision capabilities, such as performing OCR on a large batch of images, through the `curator` library's batch processing interface.

## Why is this change needed?

Previously, attempting to run a batch job with an image and text resulted in a `400 INVALID_ARGUMENT` error from the Gemini API. The root cause was that the request processor was not correctly formatting the multimodal payload. It tried to serialize the entire content dictionary into a `{"text": ...}` field, which is invalid.

The Gemini Batch API requires multimodal inputs to be structured as a list of `parts` within a single `contents` object, where each part is either text (`{"text": ...}`) or file data (`{"fileData": ...}`). This PR implements that required structure.

## How was this change implemented?

The core of the change is within the `create_api_specific_request_batch` method:
1.  **Inspects Content Type:** The method now checks if `message["content"]` is the dictionary structure that `curator` creates for multimodal prompts (e.g., `{'texts': [...], 'images': [...]}`).
2.  **Parses Multimodal Dictionary:** It iterates through the `texts` and `images` lists within the content dictionary.
3.  **Builds Correct Payload:**
    *   For each text string, it creates a `{"text": "..."}` part.
    *   For each image, it creates a `{"fileData": {"fileUri": "...", "mimeType": "..."}}` part. The `mimetype` is automatically detected from the image URI.
4.  **Assembles Parts:** All generated parts are collected into a single `parts` array, creating a valid request payload for the Gemini Batch API.

A fallback for legacy text-only requests is maintained for backward compatibility.

## How to test this change?

A reviewer can test this by running a batch job with a `curator.LLM` class that returns both text and an image from its `prompt` method.

**Example:**

```python
import curator
from datasets import Dataset

# Prerequisite: An image must be uploaded to a GCS bucket.
# For example: gs://my-bucket/images/page-1.png

class PageOCR(curator.LLM):
    """A page OCR that extracts text from an image of a page."""
    def prompt(self, input: dict) -> list:
        prompt_text = "Please perform OCR on this image."
        image_url: str = input["image_path"] # e.g., "gs://my-bucket/images/page-1.png"
        return prompt_text, curator.types.Image(url=image_url)

    def parse(self, input: dict, response: str) -> dict:
        return {"ocr_text": response}

# Configure with Gemini backend and batch=True
llm = PageOCR(
    model_name="gemini-1.5-flash-latest", # Or another vision-capable model
    backend="gemini",
    batch=True
)

# Create a dataset with GCS image paths
dataset = Dataset.from_dict({
    "image_path": ["gs://my-bucket/images/page-1.png", "gs://my-bucket/images/page-2.png"]
})

# Run the batch job
result_dataset = llm(dataset)

# The batch job should now be submitted successfully to Vertex AI without errors.
```

### DIFF
--- a/src/bespokelabs/curator/request_processor/batch/gemini_batch_request_processor.py
+++ b/src/bespokelabs/curator/request_processor/batch/gemini_batch_request_processor.py
@@ -1,5 +1,6 @@
 import copy
 import json
+import mimetypes
 import os
 import typing as t
 from functools import lru_cache
@@ -228,8 +229,8 @@ class GeminiBatchRequestProcessor(BaseBatchRequestProcessor):
         """Creates an API-specific request body from a generic request.
 
         Transforms a GenericRequest into the format expected by Gemini's batch API.
-        Combines and constructs a system message with schema and instructions using
-        the instructor package for JSON response formatting.
+        Handles multi-modal inputs by parsing the dictionary structure created by the
+        curator framework, which combines text and images into a single message.
 
         Args:
             generic_request: The generic request object containing model, messages,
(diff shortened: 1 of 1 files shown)


kartik4949:
@alexisdrakopoulos Thanks for the contribution,
Could you add a test for this?

src/bespokelabs/curator/request_processor/batch/gemini_batch_request_processor.py
```diff
@@ -239,17 +240,57 @@ def create_api_specific_request_batch(self, generic_request: GenericRequest) ->
...
+        parts = []
+        # The prompt() function returns (text, image), which curator combines into
+        # a single message with a dictionary as its content. We need to parse this dict.
         for message in generic_request.messages:
-            contents.append({"role": message["role"], "parts": [{"text": message["content"]}]})
-        request_object = {"contents": contents}
+            content = message["content"]
+
+            # Primary Path: Handle the multimodal dictionary from curator
+            if isinstance(content, dict) and ("texts" in content or "images" in content):
```
kartik4949, line 250: do you need the "texts" or "images" check?

src/bespokelabs/curator/request_processor/batch/gemini_batch_request_processor.py
```diff
@@ -239,17 +240,57 @@ def create_api_specific_request_batch(self, generic_request: GenericRequest) ->
...
+                # Add all text parts from the 'texts' list
+                for text_item in content.get("texts", []):
+                    if isinstance(text_item, str):
+                        parts.append({"text": text_item})
+
+                # Add all image parts from the 'images' list
+                for image_item in content.get("images", []):
```
kartik4949, line 257: there are "files" as well

src/bespokelabs/curator/request_processor/batch/gemini_batch_request_processor.py
```diff
@@ -239,17 +240,57 @@ def create_api_specific_request_batch(self, generic_request: GenericRequest) ->
...
+                    url = image_item.get("url")
+                    if not url:
+                        continue  # Skip if there's no URL
+
+                    mime_type = image_item.get("mime_type")
+                    if not mime_type:
+                        # Fallback to guessing the mime type if not provided
+                        mime_type, _ = mimetypes.guess_type(url)
+                        if not mime_type:
+                            logger.warning(f"Could not determine MIME type for {url}. Defaulting to 'image/png'.")
```
kartik4949, line 268: logger.debug

src/bespokelabs/curator/request_processor/batch/gemini_batch_request_processor.py
```diff
@@ -239,17 +240,57 @@ def create_api_specific_request_batch(self, generic_request: GenericRequest) ->
...
+                parts.append({"text": str(content)})
+
+        # For multi-modal requests, Gemini expects a single 'contents' object
+        # with a 'parts' list. The role is typically 'user'.
+        request_object = {"contents": [{"role": "user", "parts": parts}]}
+
         if generic_request.response_format:
-            request_object.update(
+            # Ensure generationConfig exists before trying to update it
+            if "generationConfig" not in request_object:
```
kartik4949, line 288: we are creating `request_object` with "contents" key only
how will this if statement hit?

kartik4949 (changes requested):
Looking good, lets do the required changes and add an integration test for this
Thanks

src/bespokelabs/curator/request_processor/batch/gemini_batch_request_processor.py
xingyi-Starry: I don't think it's a good idea to only log debug-level messages when the mime_type is missing, as this makes it hard for people to notice the problem—even though other API implementations might not log anything at all. I believe that at the very least a warning should be issued when mime_type is missing, or even an exception should be thrown.

PR #698, opened 2025-10-21
fix: ensure proxy support and bump anthropic dependency

jinzhuer:
### Root cause

1. Curator’s aiohttp.ClientSession was created without trust_env=True, so it ignored HTTP(S)_PROXY.

2. Older anthropic SDK versions didn’t provide robust proxy support, so Anthropic calls didn’t honor the proxy.
### Changes
1. Injected `trust_env=True)` when creating the `aiohttp.ClientSession`, ensuring system proxy variables (`HTTP(S)_PROXY`, `ALL_PROXY`) are respected.

2. Bumped the `anthropic` dependency to a newer version that has native proxy support.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -202,28 +202,30 @@ files = [
 
 [[package]]
 name = "anthropic"
-version = "0.47.2"
+version = "0.71.0"
 description = "The official Python library for the anthropic API"
 optional = false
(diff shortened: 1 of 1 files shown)


jinzhuer:
fix issue #699

PR #730, opened 2026-07-16
Return cached partial results after interrupt

### LINKED ISSUE #618: Question/Proposal: Option to end sending requests gracefully and retain successful requests
Use case:
Some APIs, like DeepSeek R1 and OpenAI, can get busy and don't respond to requests after some time. I want an option to gracefully stop sending requests and also retain data of successful requests. 

I'm looking through .cache/curator and I only see an .arrow file associated with a **successful** request load. It would be nice if:
- We can download data from curator viewer or
- Data of successful requests are added to a database as they're done, instead of waiting on a long request

RitwijParmar:
Closes #618

This adds an opt-in path for long Curator runs where Ctrl+C should not throw away useful completed work.

What changed:
- Adds `allow_partial_result_on_interrupt` to backend params.
- Keeps the old behavior by default. If the flag is not set, `KeyboardInterrupt` is re-raised.
- When the flag is set, Curator builds the output dataset from successful responses already written to cache.
- Unfinished requests are still written to `failed_requests.jsonl`, so the partial run is inspectable and resumable.
- Documents the flag in the README.

Why this matters:
For large online or batch-backed generation jobs, users often stop a run because the prompt is wrong, the provider is slow, or the dataset is larger than expected. The successful responses are already durable in Curator cache. This makes that path explicit instead of making users recover it by hand.

Local checks:
- `.venv/bin/python -m pytest tests/unittests/test_partial_result_interrupt.py tests/unittests/test_base_online_request_processor.py -q`
- `.venv/bin/ruff check src/bespokelabs/curator/request_processor/config.py src/bespokelabs/curator/request_processor/base_request_processor.py tests/unittests/test_partial_result_interrupt.py`
- `.venv/bin/ruff format --check src/bespokelabs/curator/request_processor/config.py src/bespokelabs/curator/request_processor/base_request_processor.py tests/unittests/test_partial_result_interrupt.py`

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,6 +190,16 @@ for troubleshooting information.
 > So now if you run the same prompt again, you will get the same response, pretty much instantly.
 > You can delete the cache at `~/.cache/curator` or disable it with `export CURATOR_DISABLE_CACHE=true`.
 
+For long runs, you can opt into partial results after a keyboard interrupt:
+
+```python
+llm = curator.LLM(
+    model_name="gpt-4o-mini",
+    backend_params={"allow_partial_result_on_interrupt": True},
+)
+```
+
+If you press Ctrl+C, Curator returns the successful responses already written to cache and records unfinished requests in `failed_requests.jsonl`.
 
 > [!IMPORTANT]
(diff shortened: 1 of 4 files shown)


RitwijParmar:
The useful next step here is making partial results explicit enough to continue the run rather than only inspect it. I would add a resume manifest with completed request ids, failed request ids and the backend parameters used for the interrupted batch. A later run could consume that manifest and skip work that already succeeded. For long generations that changes an interrupt from lost time into a recoverable checkpoint.

